PR #23, opened 2021-09-29
Jd 2021 09 django 3 webpack

goapunk:
* Upgrade to Django 3.2 as the new LTS release
* Fix an error in the webpack.prod config
* Update webpack
* Remove unusued and unmaintained packages

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -1,95 +1,133 @@
 const webpack = require('webpack')
 const path = require('path')
+const glob = require('glob')
 const CopyWebpackPlugin = require('copy-webpack-plugin')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 
 module.exports = {
   entry: {
-    adhocracy4: [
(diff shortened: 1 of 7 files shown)


webpack.common.js
```diff
@@ -1,95 +1,133 @@
...
   module: {
     rules: [
       {
         test: /\.jsx?$/,
-        exclude: /node_modules\/(?!(bootstrap)\/).*/, // exclude all dependencies but bootstrap
-        loader: 'babel-loader',
-        options: {
-          presets: ['@babel/preset-env',].map(require.resolve),
-          plugins: ['@babel/plugin-transform-runtime', '@babel/plugin-transform-modules-commonjs']
+        exclude: /node_modules\/(?!(bootstrap)\/).*/, // exclude all dependencies but adhocracy4
```
goapunk, line 41: do we actually still need this?

fuzzylogic2000: This is not outdated, I only changed the comment...

PR #24, opened 2021-09-30
jd 2021 09 add ci config

goapunk:


fuzzylogic2000 (commented):
Looks good!
But we do not use some stuff from the makefile here. We don't do transifex (`tx-mo`) and don't have background tasks and I think we also will not add them. So far we also don't have fixtures and for the other pure wagtail sites we also didn't write tests. Those would be nice to have at some point, though.
Should we merge and remove/add later or talk about it?

goapunk:
I can remove the unused stuff, no need to carry that around then! I'd just keep the test stuff?

fuzzylogic2000:
> I can remove the unused stuff, no need to carry that around then! I'd just keep the test stuff?

And the fixture! It would be cool to add some for all the blocks and so on we have, so getting back into the project will be easier!

fuzzylogic2000:
Actually I will do that! Would like to merge it now. :)

PR #25, opened 2021-09-30
README: added superuser steps

sabinammm:


README.md
```diff
@@ -13,7 +13,15 @@ Start server at port 8000
 ```
 $ make watch
 ```
-Run the tests
+### Create superuser to administer the page:
 ```
-$ make test
+$ python manage.py createsuerpuser
```
fuzzylogic2000, line 18: The p is at the wrong place. :)
And adding `source venv/bin/activate` would also be good!

sabinammm: Ups, sorry ...fixed it:)

README.md
```diff
@@ -9,11 +9,21 @@ Install dependencies
 ```
 $ make install
 ```
-Start server at port 8000
+### Start virtual environment:
+    source venv/bin/activate
```
fuzzylogic2000, line 13: You don't need to add it before the make watch. That works without starting the venv, because it is using the python from the venv folder.
https://github.com/liqd/digitalstrategie/blob/89717a0da2283fc5fc410e9e02eaecf28fc59d5c/Makefile#L49

I thought, you just add it before `python manage.py createsuperuser` because you cannot use the python there without starting the virtual env.

fuzzylogic2000 (commented):
I changed it a bit if you are happy with it like that you can merge @sabinammm

sabinammm (commented):
Yes, thankk you :)

PR #28, opened 2021-10-01
apps/home: added faq block to home and simple page

sabinammm:
@fuzzylogic2000 The accordion (a4ckeditor --> collapsible_fragment.html) isn't styled properly and doesn't open up. Should we merged once it's tested it anyway, so that @Rineee  can keep working on the model?


fuzzylogic2000:
> @fuzzylogic2000 The accordion (a4ckeditor --> collapsible_fragment.html) isn't styled properly and doesn't open up. Should we merged once it's tested it anyway, so that @Rineee can keep working on the model?

Yes, I think that's a good idea! But @Rineee should review and merge then!

And I think @phillimorland was working on making the bootstrap js working? So, make accordeons and dropdowns and stuff like that behave like thexy should.

Rineee (approved):
Yes, nice! I'll merge an go on from here

PR #29, opened 2021-10-01
apps/snippets: add dropdown items to menu snippet and restrict menu i…

fuzzylogic2000:
…tem number

This will break your local db, so only works with a fresh one. Before merging we have to think about what to do to the dev server,


apps/snippets/models.py
```diff
@@ -24,24 +51,34 @@ def __str__(self):
...
 
     panels = [
+        FieldPanel('title'),
         PageChooserPanel('link_page'),
-        FieldPanel('title')
+        StreamFieldPanel('subpages')
     ]
 
 
+class NavigationMenuItem(Orderable, MenuItem):
```
fuzzylogic2000, line 60: @Rineee I find it weird that we have two models in their own tables here. I think either the `MenuItem` should be abstract or the `Orderable` and `parent` could just be in the `MenuItem`.
Buuut it's a bit late for that now unless I change the very first migration and then we would not only need to delete our local DBs, but also redo the one on the dev-server. Is that worth it or do we just keep the two models in two tables?

Rineee: @fuzzylogic2000 Oh yes, you are right, that is not needed! I like the solution with the abstract class. So from my side I would be in favour of removing the table, always better to clean up and at this point no one should have much data anyways.. but if you think its too much effort, I am of course also fine with it!

fuzzylogic2000: Alright, I did that! So before merging we have to think about what to do to the dev site then. :)

fuzzylogic2000:
Right, we also need translatable link titles! Will make it WIP and change when @Rineee s PR is merged

fuzzylogic2000:
> Right, we also need translatable link titles! Will make it WIP and change when @Rineee s PR is merged

Alright! The translated titles are now in the first commit. You can now try it out and review (on a fresh DB)! @phillimorland @sabinammm @Rineee @khamui 
Would be cool to merge quickly, because then we can drop the DB again and fix dev once and for all!

Rineee:
@fuzzylogic2000 I added two menu items with one submenu item, but that only shows as an empty dropdown.. did I do sth wrong or is that not ready because of missing styling?

fuzzylogic2000:
> @fuzzylogic2000 I added two menu items with one submenu item, but that only shows as an empty dropdown.. did I do sth wrong or is that not ready because of missing styling?

Noooo, I need to check that! Forgot about the translations there in the html, sorry.

apps/snippets/models.py
```diff
@@ -1,47 +1,101 @@
...
+    parent = ParentalKey('apps_snippets.NavigationMenu',
+                         related_name='menu_items')
+
+
```
sabinammm, line 81: Since I'm getting AttributeError at /admin/pages/add/apps_home/simplepage/3/
'NoneType' object has no attribute '_inc_path' I think that this may not be working as it should. But I don't know that this is helpful information

sabinammm (commented):
I tested this with a new db. I can add a navigation menu, but it doesn't show up, and except for the error in the simple page, I don't have any additional information.

apps/snippets/models.py
fuzzylogic2000: @sabinammm That error is solved here: https://github.com/liqd/digitalstrategie/pull/37 (hopefully!).

fuzzylogic2000:
@Rineee It's finally fixed! Can you have another look?

Rineee (approved):
Yeah! Supernice!

PR #30, opened 2021-10-01
apps/blocks & apps/home: make background color of faq block choosable

sabinammm:
This requires

- [x] a migration that hasn't been added because first the following questions have to be resolved:
- [x] which colors should be choosable?
- [x] where is the styling for the background-color of the accordion supposed to go?

@fuzzylogic2000 I checked out opin and a few resources, but couldn't get past the last question on my own. Grateful for any hint :)


phillimorland:
@sabinammm as we are using bootstrap accordions anyway we can get alot of dafault styling from bootstrap if we use their class names so I would use the example as a base, also means we can have stacked accordions without extra styling: 
https://getbootstrap.com/docs/5.0/components/accordion/
any additional styling you can add a component file in assets and call it faq_block or something?

fuzzylogic2000:
For the colours, I would use the ones from zeplin: https://app.zeplin.io/project/5c10cb93ddaf5a24436c79d2/screen/6155dd0abf7c77b307688b70

So, `#aca4d1`, `#f8ed45` and `#f5b4cb`.
@phillimorland any ideas about the naming? As we don't yet know what will be primary and secondary maybe we call them `$brand-purple`, `$brand-yellow` and `$brand-pink`?

phillimorland:
sorry, forgot to respond, I would just update the defaults hex for `yellow` `pink` `purple` in variables and we can assign primary and secondary later, I would like to check with Marcelo first

sabinammm:
@phillimorland thanks for all your input! I finally managed to get the styling in the template to interact with the template...
BUT the styling is still a bit strange, as there is a big gap between the first and following elements of the faq. Could you check out the faq_block.html when you get the chance?
@fuzzylogic2000 I also feel like the accordion should probably be more one block with the background color picked for the entire piece, like in OPIN:
image
rather than what I did, which creates individual pieces each with a different background like so:
image
What do you think?

apps/home/templates/apps_home/blocks/faq_block.html
```diff
@@ -3,35 +3,39 @@
...
-            </a>
+        {% if self.background_color == "yellow" %}
+        <div class="accordion yellow">
+            {% elif self.background_color == "pink"%}
+            <div class="accordion pink">
+                {% elif self.background_color == "purple"%}
+                <div class="accordion purple">
+                    {% else %}
+                    <div class="accordion">
+                        {% endif %}
```
fuzzylogic2000, line 14: The indentation for the ifs until the else should be the same. And is the colour choice required? In that case you do not need an `else` or can use it for your favourite colour.

fuzzylogic2000: Ah, you use it for white? So, when it is not set it's white? I think that info could go into a help text for the colour field.

digitalstrategie/assets/scss/components/_faq_block.scss
```diff
@@ -0,0 +1,13 @@
+.accordion {
+    background-color: #fff;
+
+    &.yellow {
+        background-color: #f8ed45;
```
fuzzylogic2000, line 5: You shouldn't use the colours in the componnets, but have a variable for the colours. @phillimorland suggested to replace pink, purple and yellow in the variables file (https://github.com/liqd/digitalstrategie/blob/20549143f4e76e397fa07400fd43d2b4166dd527/digitalstrategie/assets/scss/_variables.scss#L12) and use these here.

apps/home/templates/apps_home/blocks/faq_block.html
fuzzylogic2000: Hmm, it's not working to not set a colour with a dropdown, right? So you need to add the white as option.

sabinammm: If I add white as an Option, then I need another migration. Right now the buttons are "Pink", "Purple", "Yellow", and "----".

sabinammm: I changed it, but I left the else for white (it's not my favorite, but I figured as it's the default, it makes the most sense) :)

apps/home/templates/apps_home/blocks/faq_block.html
```diff
@@ -1,39 +1,33 @@
...
+                        <div class="card-body">
+                            {{ item.body }}
+                        </div>
+                    </div>
+                </div>
+                {% endfor %}
+            </section>
+        </section>
+    </section>
+</section>
```
fuzzylogic2000, line 33: Here should be only one closing section. With the if... there is also only one openend. And when that's fixed, maybe the auto-indentation will also work? Because that is still wrong for the if block.

sabinammm: Yes, that's what I thought. But when I take these off, it shows me that it's an error. I'll ignore it and change it, though, because that makes a lot more sense....

digitalstrategie/assets/scss/components/_faq_block.scss
```diff
@@ -0,0 +1,13 @@
+.accordion {
+    background-color: $white;
```
fuzzylogic2000, line 2: I guess, if you leave out the background colour out, it is also white? I think instead of wetting it white here, we should so that, so that we use the background colour we use elsewhere. (So that if we use a light gray or something that is not overwritten)

apps/blocks.py
```diff
@@ -41,10 +41,22 @@ class Meta:
...
+class AccordionItemBlock(blocks.StructBlock):
     title = blocks.CharBlock()
     body = blocks.RichTextBlock(required=False)
 
+
+class FaqBlock(blocks.StructBlock):
+
+    accordion_items = blocks.ListBlock(AccordionItemBlock())
+
+    background_color = blocks.ChoiceBlock(required=False, choices=[
```
fuzzylogic2000, line 53: Sorry, I wasn't more clear here earlier. I think you either need the `----` (the empty choice) or the white choice. If you go with the white, you should make this field requiered, so that the empty choice is gone. If you'd rather have the empty choice you could add a helptext like "Not choosing a colour will result in a FAQ block with the background of the page."

fuzzylogic2000 (commented):
Yeah! It works well! 
Just a few more remarks... 
And I found something cool, so that you can make the colour-choice block reusable: https://docs.wagtail.io/en/latest/reference/streamfield/blocks.html?highlight=ChoiceBlock#field-block-types (go down to choice block). I think that would be great as we will use it for more blocks. (and if you do that, I think my prefered option would be having the empty choice and the helptext because that works with every background colour)

digitalstrategie/assets/scss/components/_faq_block.scss
```diff
@@ -0,0 +1,13 @@
...
+
+    &.yellow {
+        background-color: $yellow;
+    }
+    &.pink {
+        background-color: $pink;
+    }
+    &.purple {
+        background-color: $purple;
```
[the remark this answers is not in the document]
sabinammm: I don't understand this: I copied these variables out of the _variables.scss. What am I not understanding...?

digitalstrategie/assets/scss/components/_faq_block.scss
sabinammm: yes, will do

apps/home/templates/apps_home/blocks/faq_block.html
```diff
@@ -1,39 +1,30 @@
...
+<section class="panel-group accordion purple accordion-content accordion-block block" role="tablist" aria-multiselectable="true">
+{% else %}
+<section class="panel-group accordion accordion-content accordion-block block" role="tablist" aria-multiselectable="true">
                 {% endif %}
-            </div>
-        </div>
-    </div>
-</div>
-
+                {% for item in self.accordion_items %}
```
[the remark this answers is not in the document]
sabinammm: +1

apps/home/templates/apps_home/blocks/faq_block.html
```diff
@@ -1,39 +1,30 @@
...
+                <div class="card card-default">
+                    <div class="card-header" role="tab" id="headingOne">
+                        <h3 class="accordion-title">
+                            <a class="accordion-link collapsed" role="button" data-bs-toggle="collapse" data-parent=".accordion-block" href="#collapse{{ forloop.counter }}" aria-expanded="false" aria-controls="collapse{{ forloop.counter }}">
+                                {{ item.title }}
+                                <i class="fa fa-plus" aria-hidden="true"></i>
+                            </a>
+                        </h3>
+                    </div>
+                    <div id="collapse{{ forloop.counter }}" class="panel-collapse collapse" role="tabpanel" aria-labelledby="heading{{ forloop.counter }}">
```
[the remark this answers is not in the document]
sabinammm: +1

digitalstrategie/assets/scss/style.scss
```diff
@@ -1,4 +1,4 @@
-// Required
+// Requireddi
```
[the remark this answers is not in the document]
sabinammm: +1

sabinammm:
@fuzzylogic2000 I noticed that the helptext isn't true unless I set the ColorChoiceField to required=false because choosing "----" will result in an error, as this choice does actually not constitute a filled in required field. I.e. it seems to be behaving differently than before.

fuzzylogic2000 (approved):
Nice! No more remarks. :)

PR #31, opened 2021-10-01
Ks 2021 09 add languages

Rineee:
I wasnt sure about changing the image model that it used on the home page, because we are not sure if we use it anyway? That is why I added that migration separately, so we could also remove it again..


apps/contrib/translations.py
```diff
@@ -1,73 +1,29 @@
...
-    def _get_model_fields_spec(self):
-        translated_attrs = {}
-        for lang_code, lang_name in LANGUAGES:
-            if lang_code == self.fallback_lang and self.overwrite_fallback:
-                name, path, args, kwargs = self.field.deconstruct()
-                kwargs.update(self.overwrite_fallback)
-                field = self.field.__class__(*args, **kwargs)
+    def hasContent(self, field):
+        if isinstance(field, StreamValue):
+            return len(field) != 0
```
Rineee, line 13: This line I had to adjust, because apparently there is a change in wagtail 2.14, see [here](https://githubmemory.com/repo/infoportugal/wagtail-modeltranslation/issues/346). I then copied from their code [here](https://github.com/infoportugal/wagtail-modeltranslation/blob/c0c3788628b9d6c5bb3f6c3d06dbfd483c2b2a2a/wagtail_modeltranslation/patch_wagtailadmin.py#L578) and this works now. We might have to do that also in other projects when we update wagtail..

fuzzylogic2000 (approved):
Yeah! Works! I think we just keep the image for now and can still remove it when we don't need it.

fuzzylogic2000 (commented):
Noooo! I didn't test this on an empty db and that is failing. I think, because we have migrations to add the homepage (which need to stay) and it's doing the images before? Anyway, I think we have to either fix this on main or undo the merge?

PR #32, opened 2021-10-04
Pm 2021 10 bootstrap optimisation

phillimorland:


phillimorland:
this seems to work with the limited amount of styling we have already but the fonts are a little funny but didn't want to faff with that as we have to load one anyway

fuzzylogic2000 (approved):
Nice! Looks good and I don't think we need any of the a4 stuff.

PR #36, opened 2021-10-04
Ks 2021 10 add language easy german

Rineee:
Maybe wait for @fuzzylogic2000 s PR to be merged and then also add easy german to link texts here?


Rineee:
I'll make it WIP as wating for decsiosion on 'leichte' or 'einfache' Sprache.

digitalstrategie/settings/base.py
```diff
@@ -111,9 +112,22 @@
...
+    ('de-ls', _('Easy German'))
 ]
 
+# adding language info for easy german
+EXTRA_LANG_INFO = {
+    'de-ls': {
+        'bidi': False,
+        'code': 'de-ls',
+        'name': 'Easy German',
+        'name_local': u"Leichte Sprache",
```
goapunk, line 125: do we really need the `u` ?

fuzzylogic2000 (commented):
Nice!

digitalstrategie/settings/base.py
Rineee: I found somewhere that it needs to be unicode and in Aplus we also did it like that.. but to be honest not sure if that is needed, it also works without.. should I remove it you think?

Rineee:
Ok, I think we could merge this now as the name_locale could be changed easily afterwards ..?

digitalstrategie/settings/base.py
goapunk: I don't really know about it in a Django context, but afaik it was required in python2 - in python3 this become unnecessary, so yes, if it works I think we should remove it

Rineee: Done!

fuzzylogic2000 (approved):
Yeah!

PR #38, opened 2021-10-04
Pm 2021 10 header styling

phillimorland:


digitalstrategie/assets/scss/_base.scss
```diff
@@ -1,10 +1,12 @@
...
+    // scroll-padding-top: 4rem; might need
 }
 
 //this is to position the footer at bottom of page with little content, the 100px is the height of the footer
-.base__wrapper {
-    min-height: calc(100vh - 150px);
-}
+// .base__wrapper {
+//     min-height: calc(100vh - 150px);
+// }
```
phillimorland, line 12: we may need this but also might not so leaving it commented out for now

fuzzylogic2000 (approved):
Supercool!

PR #47, opened 2021-10-06
templates/apps_home: added image caption to header image

sabinammm:


fuzzylogic2000 (approved):
It's showing the caption! (We probably need to change it to show on the image later.)

PR #48, opened 2021-10-06
WIP gruenbuch detail page

khamui:


khamui:
@phillimorland @fuzzylogic2000 @Rineee @sabinammm 

before going into further work, i am strongly doubting that the approach i took is the right one and even if its working, it seems not very elegant to me. Happy to get some input from python/django/wagtail experts :) 

- loading all information into context
- using same template.html for both models (assuming both models have to carry the same properties) --> weird

apps/gruenbuch/models.py
```diff
@@ -10,17 +10,23 @@
 from apps.contrib.translations import TranslatedField
 
 class GruenbuchIndexPage(Page):
-    template = 'gruenbuch_index_page.html'
+    template = 'gruenbuch_page.html'
```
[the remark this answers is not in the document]
fuzzylogic2000: I think that is better than having the same template twice or extending!

apps/gruenbuch/models.py
```diff
@@ -1,3 +1,82 @@
...
+    subpage_types = ['apps_gruenbuch.GruenbuchDetailPage']
+
+class GruenbuchDetailPage(Page):
+
+    page_blocks = [
+        ('paragraph', blocks.RichTextBlock()),
+        ('faq_accordion', apps_blocks.FaqBlock())
+    ]
+
+    page_title_de = models.CharField(max_length=120, blank=True)
```
fuzzylogic2000, line 36: I think we should require the page_title_de and body_de, as German is the fallback language and the page doesn't make sense without it.
For subtitle, we can then ask `{% if page.subtitle %}` and only display it in that case.

apps/gruenbuch/migrations/0002_auto_20211006_1029.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.2.7 on 2021-10-06 10:29
```
fuzzylogic2000, line 1: With migrations, we try to make as few as possible, but if it helps the understanding and cleaner commits, then it's also fine to have more than one in one PR.
And we name them after what they do. You can just change the name between the `number_` and `.py`. So, you could call this `0002_add_context.py`. 
(But on the other hand wagtail is always producing lots of migrations, where it's often not clear, why they are there. So, here you can also just not do the above things. :smile: )
And remember: when you change your migrations afterwards, you need a new DB (or should make a backup of your DB before you run migrations you might want to change). As soon as migrations are merged, they shouldn't be changed anymore.

apps/gruenbuch/models.py
```diff
@@ -10,17 +10,23 @@
...
     def get_context(self, request):
         context = super().get_context(request)
-        context['gruenbuch_sections'] = GruenbuchDetailPage.objects.live().public()
+        context['book'] = self
+        context['chapters'] = self.get_children().specific()
+        context['active_chapter'] = context['chapters'].first()
+        context['next_chapter'] = context['active_chapter'].get_next_sibling()
+        if self.get_next_sibling():
+            context['next_chapter'] = self.get_next_sibling().specific
+
```
fuzzylogic2000, line 23: Instead of adding stuff to the context, you can also use properties, which can then be used like the fields of the page. Maybe that's nicer here.
```
@property
def chapters(self):
    return self.get_children().specific()
```

apps/gruenbuch/templates/gruenbuch_page.html
```diff
@@ -0,0 +1,62 @@
...
+            {% endif %}
+          </div>
+          <br />
+        {% endfor %}
+      </nav>
+      <div class="col">
+        <h1>
+          {{ active_chapter.page_title }}
+        </h1>
+        <h3>
```
fuzzylogic2000, line 37: Shouldn't this be h2 for a11y? And the we can still style it however we like. (But not sure about the a11y here.)

digitalstrategie/assets/scss/components/_books.scss
```diff
@@ -0,0 +1,15 @@
+.chapter-link {
+  & a {
+    color: $text-gray;
+    text-decoration: none;
+  }
+}
+
+.chapter-link--active {
```
fuzzylogic2000, line 8: Hmm, to be a variant, the form without the variant should be used with it. But that's a bit annoying here as they only share the `text-decoration: none;`. I would be alright with keeping it like that for lack of better naming ideas.

apps/gruenbuch/models.py
```diff
@@ -0,0 +1,128 @@
...
+    def chapters(self):
+        return self.get_children().specific()
+
+    @property
+    def active_chapter(self):
+        return self.chapters.first()
+
+    @property
+    def next_chapter(self):
+        if self.get_next_sibling():
```
fuzzylogic2000, line 31: Why does the index page have a next sibling? And if it does, that could be any page, right? So, we don't want to return that? 
(and did you check using `active-chapter.get_next_sibling` in the template? Then you don't need this property here and also not on the gruenbuch-detail.)

fuzzylogic2000: Oh, or does using `get_next_sibling` in the template not work because you need the `specific` for the property in the detail?

PR #49, opened 2021-10-06
apps/home: added detail page to homepage

sabinammm:


Rineee (changes requested):
I think what @fuzzylogic2000 meant was to first rename the SimplePage in DetailPage (so Django keeps the tables and existing  SimplePages are kept (as DetailPage :))  and then introducing a new SimplePage without blocks, but only a RichtTextField. Does that make sense?

PR #52, opened 2021-10-06
WIP Ks 2021 10 contact form

Rineee:
Totally unstyled :sweat_smile: 

@fuzzylogic2000 I thought I would also add the participation form in this PR, because Im afraid I might have to change sth in the FormPage class for it and then we wouldnt need another migration.. what do you think?

But nevertheless it would be cool if you could already have a look at it!


apps/forms/models.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class FormField(AbstractFormField):
+    page = ParentalKey('FormPage',
+                       on_delete=models.CASCADE,
+                       related_name='form_fields')
+
+
+class FormPage(AbstractEmailForm):
+    header_de = models.CharField(
```
fuzzylogic2000, line 23: more a general question, but shouldn't we always require the German fields? At least the ones we need to make the page work.

fuzzylogic2000 (commented):
The code looks fine and it works well!
Good idea, to also add the other form!

apps/forms/models.py
Rineee: Oh yes, true, I changed that to be on its own page, as thats what the designs propose and and empty page would not make sense.. I'll change it to be required then!

PR #67, opened 2021-10-11
gruenbuch/faq-block/template: duplicating list template and adjust fo…

khamui:
…r gruenbuch

### DIFF
--- a/apps/home/templates/apps_home/blocks/gruenbuch_faq_block.html
+++ b/apps/home/templates/apps_home/blocks/gruenbuch_faq_block.html
@@ -3,7 +3,7 @@
 {% get_identifier value as identifier %}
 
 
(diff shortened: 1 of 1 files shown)


apps/home/templates/apps_home/blocks/gruenbuch_faq_block.html
```diff
@@ -0,0 +1,30 @@
+{% load wagtailimages_tags wagtailcore_tags block_tags %}
+
+{% get_identifier value as identifier %}
+
+
+<div class="py-3 {% if self.background_color %}block__{{ self.background_color }}{% else %}block__body-bg{% endif %} container-fluid px-0">
+    <div class="row g-0 faq-block__row">
```
phillimorland, line 7: do we use different colour backgrounds in the grunbook?

khamui: @Rineee that part is removed by the latest commit.

PR #73, opened 2021-10-13
base.html: add tags to load the required berlin styles direct from th…

philli-m:
…eir website. header.html: replacer header code with template and add in wagtail tags for testing

### DIFF
--- a/apps/home/templates/apps_home/blocks/cta_block.html
+++ b/apps/home/templates/apps_home/blocks/cta_block.html
@@ -1,10 +1,10 @@
 {% load wagtailcore_tags %}
 
-<div class="block {% if self.background_color %}block__{{ self.background_color }}{% else %}block__body-bg{% endif %}">
+<div class="ds-block {% if self.background_color %}ds-block__{{ self.background_color }}{% else %}ds-block__body-bg{% endif %}">
     <div class="container">
-        <div clas="row block__row">
+        <div clas="row ds-block__row">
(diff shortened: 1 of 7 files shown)


digitalstrategie/templates/base.html
```diff
@@ -11,7 +11,22 @@
...
+
         <link rel="stylesheet" type="text/css" href="{% static 'digitalstrategie.css' %}" />
+
+        <!-- Berlin theme start -->
+            <link rel="stylesheet" type="text/css" href="//www.berlin.de/i9f/r1/bundle/berlin_marketing.css">
+            <link rel="stylesheet" type="text/css" href="//www.berlin.de/i9f/r1/bundle/swiper.min.css">
+            <link rel="stylesheet" type="text/css" href="//www.berlin.de/i9f/r1/bundle/shariff.min.css">
+            <link rel="stylesheet" type="text/css" href="//www.berlin.de/i9f/r1/bundle/fontawesome.all.min.css">
+
+            <script src="//www.berlin.de/i9f/r1/bundle/jquery.min.js"></script>
```
goapunk, line 23: we also load jquery through webpack, I wonder whether we should remove either this line or the webpack loading

phillimorland: ahhh good catch, right I was actually going to remove it in the webpack as we weren't using it but as it's quite big I guess it's better to remove this one,webpack config makes it gloabally available and with all usual aliases so it shouldn't make a diffrence, i'll try and see, cheers!

PR #75, opened 2021-10-14
Pm 2021 10 template 2

phillimorland:


digitalstrategie/assets/scss/_base.scss
```diff
@@ -1,22 +1,78 @@
...
 
-//reset to be more flexible for headings
 h1 {
-    margin: 0
+    font-size: $h1-font-size;
 }
 
 h2 {
-    font-size: 1.1rem;
-    font-weight: inherit;
```
phillimorland, line 16: same as above, we can't have all h2 at a small size, it's better to add a utility or a specific class for these use cases, base should be used for ensuring defaults not for styling elements

PR #79, opened 2021-10-18
Ks 2021 10 teaser blocks

Rineee:
This adds the (translated) teaser fields to detail page and gruenbuch index page. I also added four teaser blocks, which then need to be available on overview pages (in the common panel). 2 teaser block templates are still missing and then ones there are also not really styled ...


apps/home/blocks.py
```diff
@@ -103,3 +103,94 @@ class TextWithImageBlock(blocks.StructBlock):
...
+
+    link = blocks.PageChooserBlock(
+        target_model=['apps_home.DetailPage',
+                      'apps_gruenbuch.GruenbuchIndexPage'],
+        help_text='Make sure that teaser title and teaser intro '
+                  'of the chosen page are set.'
+    )
+    link_text = blocks.CharBlock(max_length=50, label='Link Text')
+    background_color = ColorChoiceBlock(
+        help_text='Not choosing a colour will result in a CTA '
```
fuzzylogic2000, line 120: The helptext is a bit wrong, as this is no CTA block. Maybe we can come up with a helptext that works for all blocks and change it everywhere? Maybe just "Not choosing a colour will result in a block with a white background."?

Rineee:
> Yes! Looks good to me, apart from the one helptext thing, which you can either fix now and then merge or we can also merge and fix that later. :)

Ah damn, copy paste mistake :) I will fix it now for all the teaser blocks. Or do you mean for all blocks that use the background colour? @fuzzylogic2000

Rineee:
> Just for my general understanding: which design shows that the teasers are used in the gruenbuch page? I'm only finding it for the weissbuch. Everything looks great - thanks for changing the blocks helptext :) and I will get to the templates as soon as it's done.

@sabinammm It is the other way around, so the teasers should be able to link to the grünbuch index page.. you can see an example here: https://app.zeplin.io/project/5c10cb93ddaf5a24436c79d2/screen/6155dd1ec6a900ae5c82ee75

fuzzylogic2000 (approved):
Nice!

PR #80, opened 2021-10-18
WIP gruenbuch/pages/none-case: avoiding site from showing error, if there…

khamui:
… were no chapters/pages added yet to Gruenbuch Index Page


apps/gruenbuch/models.py
```diff
@@ -18,17 +18,27 @@ class GruenbuchIndexPage(Page):
...
     @property
     def chapters(self):
-        return self.get_children().specific()
+        if self.get_children():
+            return self.get_children().specific()
 
     @property
     def active_chapter(self):
-        return self.chapters.first()
+        if self.chapters:
```
fuzzylogic2000, line 35: I find it somehow weird that the properties always ask for the one before. On the other hand, they also use what they ask for, so, keep that!
But I think, they should always return something either in an else or just by indentation.
```
    @property
    def active_chapter(self):
        if self.chapters:
            return self.chapters.first()
        return ''
```
I am not sure, what a good empty value to return would be, though. The empty string could work.

apps/gruenbuch/templates/gruenbuch_page.html
```diff
@@ -9,64 +9,68 @@
...
-      <nav class="col-sm-4 col-md-3 col-xl-2">
-        {% for chapter in page.chapters %}
-          <div class="m-1 fs-6 fw-bolder">
-            {% if chapter.url == page.active_chapter.url %}
-              <div class="chapter-link chapter-link--active">
-                <i class="fas fa-arrow-right"></i>
-                <a href="{{ chapter.url }}">
-                  {{ chapter.page_title }}
+      {% if page.isEmpty %}
+        <div>{% trans 'No chapters/content has been added, yet.' %}</div>
```
fuzzylogic2000, line 13: Thie sound weird to me. Maybe no comma? And just the chapters. "No chapters have been added yet."

apps/gruenbuch/templates/gruenbuch_page.html
```diff
@@ -9,64 +9,68 @@
...
+      {% else %}
+        <nav class="col-sm-4 col-md-3 col-xl-2">
```
fuzzylogic2000: Just becasue I don't know: Without writing col-12 here, that still happens for the smallest sizes?

apps/gruenbuch/templates/gruenbuch_page.html
```diff
@@ -9,64 +9,68 @@
...
+                  <a href="{{ chapter.url }}">
+                    {{ chapter.page_title }}
+                  </a>
+                </div>
+              {% endif %}
+            </div>
+            <br />
+          {% endfor %}
+        </nav>
+        <div class="col">
```
fuzzylogic2000, line 34: And another col-question: so, this just gets what's left over? But for the smallest size also its own column?

digitalstrategie/assets/scss/components/_books.scss
```diff
@@ -1,4 +1,19 @@
+// Books (Grünbuch and Weißbuch) specific overwrites
```
fuzzylogic2000, line 1: But this overwrites all headlines, doesn't it? 
So, either add utility classes with what they do
```
.u-font-size-h2 {
    font-size: $h2-font-size;
}
```
(https://github.com/liqd/digitalstrategie/blob/40da7a3f22848688e33a98fedd1f43b85a586cc6/digitalstrategie/assets/scss/_variables.scss#L197)

Alright, while looking up the bootstrap utility for the font weight, I found that there are also some for the font sizing! So, you could replace all of the headline styling by using bootstraps utility classes directly in the template:
`fw-normal`, `fs-2`, `fs-5` and `mt-0`.

https://getbootstrap.com/docs/5.0/utilities/text/

Or instead of utilies, you could also use specific class names like `books__h1`

digitalstrategie/assets/scss/components/_books.scss
```diff
@@ -1,4 +1,19 @@
...
+}
+
+h2 {
+    font-size: $font-size-base * 1.125;
+    font-weight: normal;
+}
+
+// Books specific nav link colors
 .chapter-link {
+    font-size: $font-size-base * 1.125;
```
fuzzylogic2000, line 14: For the font-size, you should also use a variable.

digitalstrategie/assets/scss/components/_books.scss
```diff
@@ -1,15 +1,39 @@
...
+.books__wrapper {
+    background-color: $green;
+}
+
+.books__content {
+    color: $white;
+
+    & h1 {
+        color: $white;
+        font-size: $font-size-base * 2;
```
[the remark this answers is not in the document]
khamui: > Very nice! The download button is a bit hard to see, but we can fix that another time!

oh but that is important? I will make a fix PR for this and for the h1 font-size

PR #81, opened 2021-10-18
Ks 2021 10 overview page

Rineee:
So again only basic template, but I thought in case you want to go on with that, this could already be merged.. otherwise I'll go on tomorrow!

I branched this off from the teaser blocks branch https://github.com/liqd/digitalstrategie/tree/ks-2021-10-teaser-blocks, so this probably needs rebasing then after the teaser blocks are merged..


Rineee:
Ok, so definitely need rebasing before merge :)

apps/home/models.py
```diff
@@ -78,7 +78,8 @@ class HomePage(Page):
         ObjectList(de_ls_content_panels, heading='Easy German'),
     ])
 
-    subpage_types = ['apps_home.DetailPage',
+    subpage_types = ['apps_home.OverviewPage',
+                     'apps_home.DetailPage',
```
fuzzylogic2000, line 82: We could also (or only) add the detail pages as subpages to the overview page. Maybe it's easier to keep an overview about the pages then?!

Rineee: Yes, I was also wondering about that, but then wasnt sure if we would maybe still need the detail page as subpage of homepage somewhere? But adding them as subpages to overview page definitely makes sense!

PR #85, opened 2021-10-19
apps/home: added basic styling to overview page

sabinammm:
I modeled the title styling on those overview page designs that contained a title ("Section Title") because the title is currently required. I also left the teaser styling to the templates for each. Should there be additonal styling in the overview template for the teasers as there is for the blocks on the homepage?


apps/home/templates/apps_home/overview_page.html
```diff
@@ -2,11 +2,23 @@
 {% load wagtailcore_tags %}
 
 {% block title %}
-{{ page.title }}
+<div class="ds-block__header_img_overlay">
+    <div class="container">
+            <div class="justify-content-md-center">
+                <h2>{{ page.title }}</h2>
```
fuzzylogic2000, line 8: The title is wagtail's thing to have a page title and get a slug from, so while we always need it, we do not use or display it as it is not translated or translateable. 
If we have a seperate title, we always call it differently and add it as a translatable field. Either `subtitle` like for the homepage or `page_title` like for gruenbuch.

fuzzylogic2000 (commented):
Looks good! But the title should be removed.

PR #90, opened 2021-10-19
Implement full-text search with elasticsearch

Rineee:
@sabinammm So this is what I came up with so far.. using wagtail postgres search backend, then adding (most of) the page fields to search index and a very simple view that lists all search results.
To do/open questions 

- [ ] what do we do with the translated fields -> eg only search german fields on german site
- [ ] should teaser fields be added to search
- [ ] should form fields be added to search
- [ ] optimize search -> order
- [ ] style result page
- [ ] is this enough or do we need js for immediate display of results
- [ ] ...

### DIFF
--- a/apps/contrib/translations.py
+++ b/apps/contrib/translations.py
@@ -31,3 +31,14 @@ class TranslatedField(object):
             return de_ls
         else:
             return de
+
+
+def get_search_fields():
+    fields = [
+        '*page_title_',
+        '*page_intro_'
+        '*subtitle_',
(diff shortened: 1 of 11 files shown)


goapunk:
@rine I think this should be ready, I also added a very simple test. As you already know how to setup elasticsearch, could you review it ? :)

Rineee:
I think the 'no results found' never displays, but I will add an an issue for that. So alright, lets give this a go!

PR #92, opened 2021-10-20
forms/styling: contact form and participation forms affected

khamui:
Just to mention, basically we use all form stylings from CI, but on some locations we kind of need to overwrite them to match the design. Not sure of this is a good practice though.


sabinammm (approved):
Looks really nice!

digitalstrategie/assets/scss/_variables.scss
```diff
@@ -11,6 +11,7 @@ $indigo:  #6610f2 !default;
 $purple:  #aca4d1 !default; // from ds design
 $pink:    #f5b4cb !default; // from ds design
 $red:     #dc3545 !default;
+$red20:   #dc354520 !default; // added
```
fuzzylogic2000, line 14: I think, you can change the red that's already there and then just make a note that it's from the design now!

fuzzylogic2000 (commented):
Looks good, but you are right, now it's even harder to see what's our styling, what's theirs and what's from bootstrap. :grimacing: But I like your approach with the form-overwrite!

apps/forms/templates/apps_forms/includes/form.html
```diff
@@ -4,17 +4,27 @@
...
                         {% include 'apps_forms/includes/form_field.html' with field=field %}
                     {% endfor %}
                 </div>
-                <div class="mb-3">
-                    <button class="btn btn--transparent btn--full" type="submit">{% trans "Send" %}</button>
+                <div class=info-section__form>
+                    <span>{% trans '*Pflichtfelder' %}</span>
+                    {% if page.isParticipation %}
+                        <p>
+                            {% trans 'Wenn Sie ihre personenbezogen Daten löschen lassen möchten, kontaktieren Sie: <span>digitalstrategie@senweb.berlin.de</span>' %}
```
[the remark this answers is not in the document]
Rineee: No, didnt handle this, I dont remember it from the old design.. should this be the same address the form is sent to? Probably we dont know.. Otherwise we just need another field in the ParticipationFormPage for it.

digitalstrategie/assets/scss/style.scss
```diff
@@ -58,3 +58,4 @@
 @import 'components/footer';
 @import 'components/header';
 @import 'components/books';
+@import 'components/forms';
```
phillimorland, line 61: this is a total nit pick so sorry about that but as component files should reflect the block class name to which they refer I would also call the file 'form' not 'forms'

digitalstrategie/assets/scss/components/_forms.scss
```diff
@@ -0,0 +1,66 @@
+.form--overwrite {
+    & form label {
+        font-weight: normal;
```
phillimorland, line 3: I think `&` is not needed as it's just a longer way to write it as it's already nested, [here ](https://css-tricks.com/the-sass-ampersand/)is good overview of ampersand. 
Also no need for `form` throughout the whole of the file

PR #181, opened 2021-11-03
apps/home: added choosable link to text with image block

### LINKED ISSUE #162: #4899 UX/UI Test - Landing Page
**URL:** https://digitalstrategie-dev.liqd.net/admin/pages/3/edit/preview/
**important screensize:** *Both*
**Comment/Question:** *There is a CTA missing here, the idea of the Landing Page is also to promote other Overview pages meanwhile talking about the project* 

Design on Zeplin: 
Screenshot 2021-10-28 at 12 45 35

Current Status: 
Screenshot 2021-10-28 at 12 42 25

Thanks!

sabinammm:
This closes #162.

@phillimorland I attempted to change the color of the link from blue to black using $link-dark for color in the class that I'm using. But this broke the page as an undefined variable, though I thought this was bootstrap 5 (??). It now looks like this:
image
Should we leave it for now?


phillimorland:
@sabinammm you can see a list of common variables in `variables.scss` in `assets` folder or to check the entire list, view it in the `node_modules` bootstrap folder or in the docs online, when you get an unknown variable error message, try another known variable to see if it at least works to change the thing you are trying to change even if it's not correct colour ect.  Then check elsewhere for correct one, if still not created and you think it will be used again (colours should always be variables for example, other things you can decide if it's needed). You can add a new variable to `core-variables` but in this case I think you can just use $black right?!

sabinammm:
> @sabinammm you can see a list of common variables in `variables.scss` in `assets` folder or to check the entire list, view it in the `node_modules` bootstrap folder or in the docs online, when you get an unknown variable error message, try another known variable to see if it at least works to change the thing you are trying to change even if it's not correct colour ect. Then check elsewhere for correct one, if still not created and you think it will be used again (colours should always be variables for example, other things you can decide if it's needed). You can add a new variable to `core-variables`

Thank you for the explanation:) $black worked and I'll remember this note for future styling.

apps/home/templates/apps_home/blocks/text_with_image_block.html
```diff
@@ -13,6 +13,11 @@ <h2>
                         {{ value.title }}
                     </h2>
                     {{ value.body|richtext }}
+                    <div>
+                        {% if value.link %}
+                        <a class="quote-ds-block__link" href="{{ value.link.url }}">{% trans 'Read more' %}&#8594;</a>
```
fuzzylogic2000, line 18: The `&#8594;` is the arrow, right? We do normally use font awesome there instead. Just use arrow-right: https://fontawesome.com/v5.15/icons/arrow-right?style=solid But don't forget to add aria-hidden=trrue.

fuzzylogic2000 (commented):
Nice! HJust one very tiny remark!

apps/home/templates/apps_home/blocks/text_with_image_block.html
sabinammm: Awesome, thanks:) I changed it, but I used the same for quote_block so should we change that or just keep in mind that we change it next time we touch it?

fuzzylogic2000: Yes, please change it in both and add `aria-hidden="true"`.
(You always need to add either the `aria-hidden` or an `aria-label` to the font awesome images to make then accessible.)

sabinammm: Thanks! I added it to the other font awesome images as well, and I hope it's all fine now :)

apps/home/templates/apps_home/blocks/quote_block.html
```diff
@@ -7,15 +7,15 @@
                 {{ self.text }}
                     <div>
                         {% if value.link %}
-                        <a class="quote-ds-block__link" href="{{ value.link.url }}">{% trans 'Read more' %}&#8594;</a>
+                        <a class="quote-ds-block__link" href="{{ value.link.url }}">{% trans 'Read more ' %}<i class="fas fa-arrow-right" aria-hidden="true"></i></a>
```
fuzzylogic2000, line 10: One last thing: There shouldn't be a blank at the end of the translation. Did you add that for the space between the "Read more" and the arrow? Then try to add it between the trans tag and the `<i...`

sabinammm: Ok, will remember this. thanks!

khamui:
looking good!

PR #200, opened 2021-12-08
search: use es index from settings

goapunk:
fixes the 500 when searching because a non-existing index is used.


fuzzylogic2000 (approved):
I was already wondering about sentry. :)

PR #210, opened 2022-01-27
[5250] - use berlin.de grid and remove bootstraps one 

fuzzylogic2000:
First commit is like https://github.com/liqd/digitalstrategie/commit/f135d688c0f93b6a89488bcd1ac8dcd241ba0586 and a bit of https://github.com/liqd/digitalstrategie/commit/370c10d4b0394b9408a24945732243068b0ae91e 

The second commit is worse than I thought. While it is kind of fine for desktop size (apart from the error pages not being centered and the paragraphs in the simple pages not breaking to new lines), most pages totally break in smaller sizes. On the greenbook page everything vanishes (!) and for example the FAQ blocks don't work with the background colours and the berlin.de grid.


phillimorland:
@fuzzylogic2000 so the layout ids have a background colour within a print media query which it also seems to apply to their mobile query, this is why the grunbuch is invisible, so we could either request they fix it (unless that was their intention but I doubt it?!) or overwrite with our own mobile query. Kind of depends what we think is best?

fuzzylogic2000:
> @fuzzylogic2000 so the layout ids have a background colour within a print media query which it also seems to apply to their mobile query, this is why the grunbuch is invisible, so we could either request they fix it (unless that was their intention but I doubt it?!) or overwrite with our own mobile query. Kind of depends what we think is best?

@phillimorland OMG! I think that is for @hmkay to decide. But I would say: ask them to fix it and if they don't, then we use our own styles? (@hmkay So, the problem here is, that the grünbuch pages vanish on mobile.)

fuzzylogic2000:
@phillimorland Is it the same problem for the other pages where the blocks look weird for smaller sizes? And should we merge nonetheless and deal with that later/when we talked to them?

phillimorland:
@fuzzylogic2000 I think thats the reason they look weird but would need to check their code closer but yes, lets merge and go one

digitalstrategie/assets/external_css/berlin_marketing.css
```diff
@@ -1,8378 +0,0 @@
...
-    z-index: auto
-}
-
-@media print, (max-width: 64rem) {
-    .layout-grid, #layout-grid {
-        display: block
-    }
-}
-
-@media print, (min-width: 64.001rem) {
```
phillimorland: @fuzzylogic2000 these are their media queries

PR #223, opened 2022-02-03
[5311] images/logo: replace old logo with new and delete png file as not used

phillimorland:


fuzzylogic2000 (commented):
It works fine in chrome, but firefox does not show the image, even though I can access it on `http://localhost:8006/static/images/logo.svg`. Should we just merge and see if firefox gets over that?

phillimorland:
yeah? I remember fff was weird when we changed logos for enquete too, i can look up why but lets see if it works after dev build?

PR #224, opened 2022-02-03
images/logo: add width to svg for firefox

phillimorland:
so in the original logo it defined the size attributes which is why it worked without scss, the new one is just the vector without any attributes. I added height and it worked in chrome as it assumes the width but not in ff, in adding dimensions this logo will also work without css which is good as means we won't have to overwrite their header


fuzzylogic2000 (approved):
Cool! And good to know!

PR #226, opened 2022-02-03
[5298] add theses block and use berlin.de styles

fuzzylogic2000:
Alright, this does it for desktop. For mobile, I shortly tried the marketing teaser styles. It didn't do anything for the colour (made it worse as was already white background on bigger screens), but might work for the spacing, which I broke earlier with my overwrite. I think, we should do the mobile-grid-redo in another PR?!


fuzzylogic2000:
@phillimorland It adds the block, uses the berlin.de styling and already overwrites some of it. But it has to have a narrower container/only use the middle columns and I am unsure how to do that. And the mobile also needs some work. Could we have a look together?

phillimorland (approved):
looking good! and yes agree we should do overhaul in a seperate PR

PR #227, opened 2022-02-03
[5309] add footer from external resource

phillimorland:


apps/snippets/models.py
```diff
@@ -95,10 +95,10 @@ class NavigationMenu(ClusterableModel):
...
         null=False,
         blank=False,
-        help_text=('Call the menu "header" to use as upper menu or '
-                   '"footer" to use in the lower dark grey footer. '
-                   'To display the upper light grey footer, call the '
-                   'menu snippet “content-footer”.')
+        help_text=('Call the menu "header" to use as upper menu. '
+                   'Call the menu “content-footer” to use '
+                   'the upper light grey footer, the lower '
+                   'footer is loaded externally.')
```
phillimorland, line 101: not sure on the wording and didn't it would trigger a migration so maybe we should check with MK first to make sure we don't have to redo?

digitalstrategie/settings/base.py
```diff
@@ -52,6 +52,7 @@
     'apps.settings',
     'apps.snippets',
     'apps.gruenbuch',
+    'digitalstrategie'
```
phillimorland, line 55: not sure if this is the best thing to do?

digitalstrategie/templatetags/extra_tags.py
```diff
@@ -0,0 +1,11 @@
...
+from django import template
+from django.utils.html import mark_safe
+
+register = template.Library()
+
+
+@register.simple_tag
+def get_external_footer():
+    response = requests.get('https://www.berlin.de/rbmskzl/__i9/ohne/foot.inc')
+    return mark_safe(response.text)
```
phillimorland, line 11: as discussed we are marking an external file safe, not idea but not sure on alternate way

fuzzylogic2000: I thought about and think it is fine. Also, there is no other way, I can think of.

digitalstrategie/templatetags/extra_tags.py
```diff
@@ -0,0 +1,11 @@
+import requests
+from django import template
+from django.utils.html import mark_safe
+
+register = template.Library()
+
+
+@register.simple_tag
+def get_external_footer():
+    response = requests.get('https://www.berlin.de/rbmskzl/__i9/ohne/foot.inc')
```
fuzzylogic2000, line 10: I guess, this should go into a setting? @goapunk ?

digitalstrategie/settings/base.py
fuzzylogic2000: I think, I will move it to the app where we have the other footer settings? But that would be the snippets... Or we add an contrib app. Or just move it to the home app as that is the main one. Anyway, it should be in the apps folder and I will have a look where it fits.

apps/snippets/models.py
fuzzylogic2000: The old wording also wasn't checked. But maybe it's a good idea to do some renaming. Will see if I can add a migration to rename the old content-footer to footer!

fuzzylogic2000 (commented):
Supercool! I hope, it's alright, when I change some smaller things in your PR.

fuzzylogic2000:
@phillimorland I added some smaller changes, but in the same commits, so that it can just be merged. 
I only moved your templatetag in the first commit. And in the second added a migration renaming the "content-footer" to "footer" when a "content-footer" exists.
Could you check if it is fine like that?

phillimorland:
seems all good, should we wait for decision on adding url to setting? @fuzzylogic2000 @goapunk

digitalstrategie/templatetags/extra_tags.py
goapunk: since we usually use a setting for external links (?) we could do that here too, yes

PR #228, opened 2022-02-07
[5345] forms/validation: small adjustments to match styleguide

khamui:


khamui:
sorry set to wip, because the messages themselves differ as well. Will have another look

khamui:
couldn't find where to change the wording of the error message. But fixed some other small style issues to match the styleguide.

phillimorland:
@khamui I think the error messages are set by django forms so would need to be overwritten I guess, is that right @fuzzylogic2000 ?

khamui:
ah just saw there is a different task for it. so this is ready to be reviewed then!

fuzzylogic2000:
> @khamui I think the error messages are set by django forms so would need to be overwritten I guess, is that right @fuzzylogic2000 ?

@phillimorland Yes, I think you are right. But as we only have the German wording, we can just overwrite the translation. :)

fuzzylogic2000 (approved):
Looks good! You are not working in this PR anymore and we can merge, right?

PR #229, opened 2022-02-07
[5342] forms/validation: overwrite error message to match styleguide

khamui:


khamui:
also set to wip, waiting for info if there are more wording changes tbd in this task.

fuzzylogic2000 (approved):
Nice! I think, we should merge so that the story is done and can be tested! And we can add more stuff later if needed.

And for bigger translation PRs (unnecessary in this tiny one), they are much easier to review with three commits:
1. code change / wording changes in the code
2. `make po` / makemessages
3. add translations

PR #230, opened 2022-02-07
[5299] adjusting teasers and further updating grid

phillimorland:


digitalstrategie/assets/scss/components/_ds-block-overwrites.scss
```diff
@@ -1,15 +1,15 @@
...
-.ds-block__maincontent {
-    min-height: 0 !important;
-    margin-top: 0 !important;
-    margin-bottom: 0 !important;
-}
+// .ds-block__maincontent {
+//     min-height: 0 !important;
+//     margin-top: 0 !important;
+//     margin-bottom: 0 !important;
+// }
```
phillimorland, line 6: I don'T think we need this anymore

phillimorland:
@fuzzylogic2000 I reworked the grid based on using their elements for our blocks, as such I changed alot of your grid stuff, sorry! but it's looking ok now, although mobile is still funky for the image teasers but will wait for their say so before changing

phillimorland:
I also just realised that i'm using the worng buttons but maybe better to clean this up first and then do more specific styling as theres already alot in here?

apps/home/templates/apps_home/blocks/text_with_image_block.html
```diff
@@ -1,21 +1,25 @@
...
+<article class="modul-teaser-marketing imageteaser fullwidth ds-block">
+    <div class="background-image {% if self.background_color %}ds-block__{{ self.background_color }}{% else %}ds-block__body-bg{% endif %}">
+    </div>
+    <div class="content-container fullwidth__inner">
+        <div class="image">
+            <div class="image__image">
+                <img src="{{ image.url }}" alt="{{ image.alt }}" class="textimg-ds-block__img" />
             </div>
+            <p class="image__copyright">
+              <!-- FIXME do we have copyright -->
```
fuzzylogic2000, line 13: I think we use the caption as copyright? Should that be renamed when we add the alt-text? Or maybe we can put a reminder about this fixme into the alt-text story?

digitalstrategie/assets/scss/components/_ds-block-overwrites.scss
fuzzylogic2000: No, that was just for overwriting when I used their grid all wrong. You can remove it!

fuzzylogic2000 (approved):
Sooo nice!

fuzzylogic2000:
@phillimorland Could you rebase and then merge?

PR #239, opened 2022-02-08
Pm 2022 02 changeing teasers

phillimorland:


apps/home/blocks.py
```diff
@@ -172,6 +178,8 @@ class Meta:
         template = 'apps_home/blocks/teaser_block_two_columns.html'
         icon = 'arrow-down'
 
+# this is the same as TextWithImage also link text is no longer needed
+
```
phillimorland, line 182: @fuzzylogic2000 I didn't do migrations as I think we should also clean out all the uneeded blocks, what do you think? i've added notes and what is not used anymore

fuzzylogic2000: Yes! Let's do that! Maybe, we can quickly go through the blocks with @hmkay @maxliqd and @mcastro-lqd ?

fuzzylogic2000 (approved):
Cool!

PR #243, opened 2022-02-10
Pm 2022 02 issues

phillimorland:


apps/home/templates/apps_home/blocks/teaser_block_centered.html
```diff
@@ -9,7 +10,15 @@ <h3 class="title">{{ value.link.specific.teaser_title }}</h3>
...
-            <a href="{{ value.link.url }}" class="button {% if self.background_color == 'green' %}button--negative{% endif %}">{{ value.link_text }}</a>
+            <a href="{{ value.link.url }}" class="button {% if self.background_color == 'green' %}button--negative{% endif %}">
+              {% if LANGUAGE_CODE == 'de' %}
+                {{ value.link_text_de }}
+              {% elif LANGUAGE_CODE == 'en' %}
+                {{ value.link_text_en }}
+              {% else %}
+                {{ value.link_text_de_ls }}
+              {% endif %}
+            </a>
```
phillimorland, line 21: wasn't sure how else to do as it's not an actual translated field but happy to change if someone has better idea

phillimorland: I did try that and couldn't get it working but i'm sure you have more tricks up your sleeve :p

PR #244, opened 2022-02-10
translations: remove duplicate and make messages

phillimorland:


locale/de/LC_MESSAGES/django.po
```diff
@@ -174,55 +183,33 @@ msgstr "Interner Serverfehler."
...
-
 #: digitalstrategie/templates/includes/header.html:18
 #: digitalstrategie/templates/includes/header.html:19
 msgid "To the home page"
 msgstr "Zur Startseite"
 
-#: digitalstrategie/templates/includes/header.html:19
-msgid "Digital strategy"
-msgstr "Digitalstrategie"
-
```
phillimorland, line 225: @fuzzylogic2000 I just deleted this but only from the german file, no clue why it couldn't just do it's usual thing, that cool?

PR #245, opened 2022-02-10
Pm kl 2022 teaser block translations

fuzzylogic2000:
This was the third commit of #243


Rineee:
> @Rineee This is where the `__get__` is from: https://github.com/wagtail/wagtail/blob/6382945a7b5e1a8adce832f3c8354cfe7fcb812b/wagtail/core/fields.py#L47

Thank you! :+1:

apps/home/templates/apps_home/blocks/teaser_block_image.html
```diff
@@ -15,7 +16,9 @@ <h2 class="cta-ds-block__title">
         </h2>
         {{ value.link.specific.teaser_intro }}
         <div class="pt-5">
-            <a href="{{ value.link.url }}" class="btn btn--huge btn-outline-dark">{{ value.link_text }}</a>
+            <a href="{{ value.link.url }}" class="btn btn--huge btn-outline-dark">
+                {{ value.link_text }}
+            </a>
         </div>
     </div>
     <div class="clearfix"></div>
```
phillimorland, line 24: this block will be removed

fuzzylogic2000: Ah, so I should just remove it (together with the paragraph block)? Thanks!

phillimorland: maybe? I was going to do a bit of a clean up now we have Marcelo's guide on zeplin, I just wanted to note it down as it didn't have the links done

phillimorland: but paragraph block should definitely go!

phillimorland: @fuzzylogic2000 ok looking at the zeplin thing there will be some kind of image teaser probably so leave it in, and we can clean up once designs are confirmed, what you think

fuzzylogic2000:
@phillimorland Now in the end, I only changed the en and de_ls blocks to not be required and with that could change the conditionals in the templates to show or fallback language, when the requested one does not exist.

phillimorland (approved):
super cool

PR #254, opened 2022-02-14
tempalte/home_page//text_with_image_block: change image tags to incre…

### LINKED ISSUE #164: #4899 UX/UI Test - Landing Page
**URL:** https://digitalstrategie-dev.liqd.net/
**important screensize:** *Both*
**Comment/Question:** *Do you know if Wagtail compress the images that much? I tried uploading them in different ways, but they always look a **little** bit pixelated to me. Is there any recommendation for uploading pictures on Wagtail? Specially considering retina, I think it would be good if I can provide them these files using the right recommendation if possible* 

Screenshot 2021-10-28 at 12 49 06

phillimorland:
…ase widths fixes #164

### DIFF
--- a/apps/home/templates/apps_home/home_page.html
+++ b/apps/home/templates/apps_home/home_page.html
@@ -3,7 +3,7 @@
 
 {% block content %}
 
-{% image page.header_image max-1920x560 as header_image %}
+{% image page.header_image original as header_image %}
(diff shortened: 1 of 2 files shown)


apps/home/templates/apps_home/home_page.html
```diff
@@ -3,7 +3,7 @@
 
 {% block content %}
 
-{% image page.header_image max-1920x560 as header_image %}
+{% image page.header_image original as header_image %}
```
phillimorland, line 6: I got this from Opin, think it's fine as max pic size is 10mb so load time should be slow and we don't show on mobile anyway

fuzzylogic2000 (approved):
Nice!

PR #288, opened 2022-03-16
search: improve search results

goapunk:
* subclass the elasticsearch wagtail class
* extract and append the highlights to the queryset
* display highlights in search result page


phillimorland:
looking good, we can test more on dev

PR #289, opened 2022-03-16
#5645 templates/homepage: adjust homepage styling to styleguide/new design

phillimorland:


apps/home/models.py
```diff
@@ -31,6 +31,13 @@ class HomePage(Page):
     subtitle_de_ls = models.CharField(
         max_length=120, blank=True, verbose_name=('Home page title de ls'))
 
+    headertext_de = models.CharField(
+        max_length=255, blank=True, verbose_name=('Home page subtitle de'))
+    headertext_en = models.CharField(
+        max_length=255, blank=True, verbose_name=('Home page subtitle en'))
+    headertext_de_ls = models.CharField(
+        max_length=255, blank=True, verbose_name=('Home page subtitle de ls'))
+
```
phillimorland, line 40: I would prefer to change `subtitle` to `headertitle` and name this `headersubtitle` but as we use subtitle everywhere I wasn't sure if it was worth it?

apps/home/migrations/0030_auto_20220317_0939.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 3.2.12 on 2022-03-17 09:39
```
phillimorland, line 1: Haven't renamed this as will change modal name once know if it's ok, also other story has lots of migrations so will need to redo this after rebase anyway

apps/home/models.py
Rineee: I am fine with renaming it!

phillimorland: coolio :)

digitalstrategie/assets/scss/components/_ds-block-overwrites.scss
```diff
@@ -4,6 +4,24 @@
         max-width: 60%;
         margin: auto !important;
     }
+
+    // overwrite of herounit text box
+    .ds-hero-image__overlay {
+        margin-top: -18 * $spacer !important;
+        margin-bottom: 5 * $spacer !important;
+        text-align: center;
```
khamui, line 12: somehow `text-align: center` is not applied in mobile/tablet screensizes.

digitalstrategie/assets/scss/components/_ds-block-overwrites.scss
```diff
@@ -4,6 +4,24 @@
...
+    }
+
+    .ds-hero-image__copyright {
```
khamui: caption does not appear below image on mobile screensize.
And for larger screensize the white background of the caption is quite big.

Bildschirmfoto 2022-03-21 um 10 50 21

khamui (changes requested):
migration works fine, found some style issues regarding screensizes.

digitalstrategie/assets/scss/components/_ds-block-overwrites.scss
phillimorland: Yo haven't seen that before, what browser and how wide is screen so I can replicate?

khamui: using chrome and above 1660 px width it starts to "grow"

PR #292, opened 2022-03-17
#5646 apps/*: rm teaser field from all relevant pages and add translated bo…

phillimorland:
…dy fields to overview page


apps/home/templates/apps_home/overview_page.html
```diff
@@ -19,8 +19,8 @@ <h1>{{ page.page_title }}</h1>
     {{ self.page_intro|richtext }}
 </section>
 <div id="layout-grid__area--maincontent">
-    {% for teaser in self.teasers %}
```
Rineee, line 22: @phillimorland ah, probably this is why it errored, because overview page was still using it then!

phillimorland: ahhh cool, ok will do it all in 1 and redo migrations and see

Rineee (approved):
Supercool!

PR #311, opened 2022-03-22
#5648 home/blocks//modal//teaser_column: rm old teaser col, add new teaser …

phillimorland:
…with migration


Rineee (approved):
Supercool!

Rineee:
@phillimorland The caption is shown bold, is that intended? I will merge now anyway!

phillimorland:
ooooo good catch, no not on our designs, will fix in next one

PR #312, opened 2022-03-22
show search term only once at top

goapunk:


goapunk:
@khamui I wasn't sure whether that was intentional, so I opened this. Feel free to close if it was.

phillimorland (approved):
looking good!

PR #317, opened 2022-03-23
apps/gruenbuch/blocks: re-using teaser blocks for gruenbuch - column …

khamui:
…block and tiles block


phillimorland (approved):
super cool! :)

PR #326, opened 2022-03-23
apps/gruenbuch/download: copying template snippet to new gruenbuch ov…

khamui:
…erviewpage template

### DIFF
--- a/apps/gruenbuch/templates/gruenbuch_overview_page.html
+++ b/apps/gruenbuch/templates/gruenbuch_overview_page.html
@@ -1,5 +1,5 @@
 {% extends "base.html" %}
-{% load wagtailcore_tags wagtailimages_tags %}
+{% load wagtailcore_tags wagtailimages_tags i18n %}
 
 {% block content %}
 {% image page.intro_image max-1920x560 as intro_image %}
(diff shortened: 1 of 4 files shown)


digitalstrategie/assets/scss/_variables.scss
```diff
@@ -17,7 +17,7 @@ $green:   #0e4c38 !default;
 $teal:    #20c997 !default;
 $cyan:    #0dcaf0 !default;
 $white:   #ffffff !default; // added
-$gray:    #545454 !default; // added
+$gray:    #dddddd !default; // added
```
khamui, line 20: i changed this to the same value as specified in the design. Since it is not used anywhere and there is no real alternative gray i changed it here.

apps/gruenbuch/templates/gruenbuch_overview_page.html
```diff
@@ -19,8 +19,34 @@ <h1>{{ page.page_title }}</h1>
...
 </section>
 <div id="layout-grid__area--maincontent">
-  {% for block in self.body %}
-      {{ block }}
-  {% endfor %}
+    {% for block in self.body %}
+        {{ block }}
+    {% endfor %}
+    {% if settings.apps_settings.Downloads.gruenbuch %}
+        {% with gruenbuch=settings.apps_settings.Downloads.gruenbuch %}
```
fuzzylogic2000, line 26: Can we call this variable download? The gruenbuch confused me. :woman_shrugging:

apps/gruenbuch/templates/gruenbuch_overview_page.html
```diff
@@ -19,8 +19,34 @@ <h1>{{ page.page_title }}</h1>
...
+        <h2>{% trans 'Downloads' %}</h2>
+        <div class="download__section">
+            <div class="download__section--start">
+                <h3>{{ gruenbuch.title }}</h3>
+                <p class="download-info__meta">
+                    {% trans 'Document' %} ({{ gruenbuch.file_size|filesizeformat }}) -
+                    {% trans 'Date:' %} {{ gruenbuch.created_at|date:"m/Y" }}
+                    <br>
+                    {% trans 'Document' %}: {{ gruenbuch.file|ext }}
```
fuzzylogic2000, line 35: I think, you can use `gruenbuch.file_extension` (or `download.file_extension`) here and don't need the template tag. (But good use of a template tag!)

digitalstrategie/settings/base.py
```diff
@@ -86,6 +86,9 @@
                 'django.contrib.messages.context_processors.messages',
                 'wagtail.contrib.settings.context_processors.settings',
             ],
+            'libraries': {
+                'custom_filters': 'apps.contrib.templatetags.custom_filters'
+            }
```
fuzzylogic2000, line 91: You shouldn't need to add it here. Just adding the templatetag into the templatetag folder (When there is an `__init__.py` in it as well) and registering it (as you did) should be enough!

fuzzylogic2000 (changes requested):
It looks really great, but you do not need a custom templatetag here.

fuzzylogic2000:
I did already do the changes, I asked for. So that we can do the other fixes and get the migrations in order.

apps/gruenbuch/templates/gruenbuch_overview_page.html
khamui: ah cool, yes that is much better. Didn't know that exists already :)

khamui:
> I did already do the changes, I asked for. So that we can do the other fixes and get the migrations in order.

okay, thank you!

PR #343, opened 2022-03-24
Revert "Update elasticsearch from 7.17.0 to 8.1.0"

fuzzylogic2000:
This reverts commit 80ee2c0415c33420fb871dc227321f97634eb820.

As updating did break the search, reverting that!


buildliqd:
@fuzzylogic2000 we should set this to stay on 7.x.x

PR #346, opened 2022-03-25
[#5697]  apps/home requirements settings templates:  social media tags

sabinammm:
To do:

- [ ] add a context-processor (https://docs.wagtail.org/en/stable/topics/pages.html) to make meta-tags available to all pages besides home page that use base.html


fuzzylogic2000:
> To do:
> 
>     * [ ]  add a context-processor (https://docs.wagtail.org/en/stable/topics/pages.html) to make meta-tags available to all pages besides home page that use base.html

I did solve it a bit different after all. This way it will be really easy to add the metatags to other pages if need be and it will already work!

fuzzylogic2000:
Right! migrations are currently broken. Will see if we have more in other PRs before I fix that.

apps/home/templatetags/custom_metadata_tags.py
```diff
@@ -0,0 +1,23 @@
...
+    """Use metadata from homepage if not set on current page.
+
+    Like https://github.com/neon-jungle/wagtail-metadata/blob/master/
+    wagtailmetadata/templatetags/wagtailmetadata_tags.py
+    but uses homepage meta tags if current site does not have own tags.
+    """
+    request = context.get('request', None)
+    if not model:
+        model = context.get('self', None)
+        if not hasattr(model, 'get_meta_title'):
```
fuzzylogic2000, line 19: we should also ask if `get_meta_description` exists and is filled in. But have to fix the migrations first before I can try.

fuzzylogic2000:
@Rineee @khamui @sabinammm This is ready now! So, the idea is: if the subpages do not have any meta tags added, the tags from the homepage are taken. (And if we add them to the subpages later, when the description is not filled in, we still take the tags from the homepage. That was not possible with the title, because if that is not filled in in the meta tags, it will be taken from the page, so that every page with the meta tags also always has a `get_meta_title`.)

digitalstrategie/templates/base.html
```diff
@@ -13,6 +13,10 @@
         <link rel="icon" type="image/png" sizes="96x96" href="{% static 'images/favicon-96x96.png' %}">
         <link rel="icon" type="image/png" sizes="16x16" href="{% static 'images/favicon-16x16.png' %}">
 
+        {% block meta_tag %}
```
khamui, line 16: a couple of lines above we still have the "static" meta title and empty meta description. Do we want to keep those?
It leads to two `<title>`, `<meta description>` etc.

khamui (commented):
pretty cool! Just wondering if the following is expected?

digitalstrategie/templates/base.html
khamui: https://github.com/liqd/digitalstrategie/blob/716b845ea8be83cbb81a9a861b7c956827d59aed/digitalstrategie/templates/base.html#L8

sabinammm: Oh, yes, this is redundant! I also noticed now, thought, that the <meta name="author" content=""> isn't rendered by the new metatags. Is this relevant to us @fuzzylogic2000 ? Otherwise, we can take out both.

fuzzylogic2000: Yes, we should take that out! Good catch!

khamui:
yeah! very nice.

PR #354, opened 2022-03-29
apps/gruenbuch/detail-page: adding template for detailpage

khamui:
@sabinammm to mention is that currently the spacing around the FAQ block is quite big. changing this would affect all blocks so i leave it for now.


fuzzylogic2000 (approved):
Nice!

PR #356, opened 2022-03-29
search: use partial_match to also mach prefixes

goapunk:
enable partial_match to also find prefixes (e.g. find ``berlin`` if you enter ``berli``)


goapunk:
I added this mostly to discuss whether we want to use this or just stick with strict search results

khamui:
oh nice! I think thats a good feature and useful.

fuzzylogic2000 (approved):
Jup, cool! Let's try on dev!

PR #374, opened 2022-03-31
templates/search: overwrite meta_tag block from base

fuzzylogic2000:
I am pretty sure this will fix https://sentry.liqd.net/organizations/liqd/issues/2046/

I used a tiny part of the code we removed when adding the meta-tags. https://github.com/liqd/digitalstrategie/pull/346/files


goapunk (approved):
cool! we'll see if it disappears

PR #376, opened 2022-03-31
deps: remove fontawesome dep

### LINKED ISSUE #372: do we use font awesome (and if yes is it still working?)
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** 
**important screensize:**
**device & browser:** 
**Comment/Question:** I just noticed that we already updated font awesome to v6. As that broke the icons on other platforms and I saw earlier, that we do not use font aweseome directly, but only via the berlin.de styles, I was wondering 
1. do we use font awesome and if not, can we remove the dependency
2. if we use it, are the icons broken?

@khamui Could you have a look?

Screenshot?

khamui:
closes #372

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
      "url" : "https://github.com/liqd/digitalstrategie.git"
    },
   "dependencies": {
-    "@fortawesome/fontawesome-free": "6.1.0",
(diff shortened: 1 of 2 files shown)


fuzzylogic2000 (approved):
Yeah!

PR #377, opened 2022-03-31
Jd 2022 03 translate search results

goapunk:


Rineee (approved):
Cool, looking good!

But because translations can sometimes behave a bit funny, when adding them, we normally do 2 commits: one with the untranslated files, so just after makemessages and then another one with the translations added.. just for next time :)

goapunk:
oh, sorry, will do next time :)

PR #415, opened 2022-05-02
[5930] apps/home: adapt quote block

fuzzylogic2000:


apps/home/templates/apps_home/blocks/quote_block.html
```diff
@@ -2,15 +2,17 @@
...
-        <p class="quote-ds-block__text">
-            <i class="fas fa-quote-left fa-xs" aria-hidden="true"></i>
-            {{ self.quote }}
-            <i class="fas fa-quote-right fa-xs" aria-hidden="true"></i>
-        </p>
-        <hr class="medium-space">
-        <p>
-            {{ self.author }}
-        </p>
+        <blockquote class="modul-testimonial">
```
phillimorland, line 5: @fuzzylogic2000 we don't need `modul-servicepanel fullwidth` and `servicepanel__main text-center` as these were there to allow the colours to be full width but when there no colours we don't need

phillimorland (commented):
looking good, would just remove the un-needed class/div

phillimorland (approved):
Coolio! :)

PR #419, opened 2022-05-03
package.json: rm un-needed deps and config and update webpack config

phillimorland:
we don't have any JS so removed the linter, the babel config could also maybe totally go but couldn't get it working so kept a barebones version.


fuzzylogic2000 (approved):
very nice!

PR #421, opened 2022-05-03
home/blocks//models//templates: add new single teaser block, update n…

phillimorland:
…aming, include link description


Rineee (approved):
Yeah, looks great!

PR #429, opened 2022-05-04
Kl 2022 05 story issues

fuzzylogic2000:


apps/home/blocks.py
```diff
@@ -11,24 +11,6 @@ class ColorChoiceBlock(blocks.ChoiceBlock):
...
-    link = blocks.PageChooserBlock(required=False)
-    link_text = blocks.CharBlock(max_length=50, label='Link Text')
-    background_color = ColorChoiceBlock(
-        help_text='Not choosing a colour will result in a '
-                  'block with a white background.',
-        required=False
-    )
-
-    class Meta:
-        template = 'apps_home/blocks/cta_block.html'
```
phillimorland, line 28: should we also remove template and any custom css?

phillimorland (commented):
just one question otherwise looking good!

apps/home/blocks.py
fuzzylogic2000: yes, totally! Forgot about that and will do!

phillimorland (commented):
perfect!

PR #430, opened 2022-05-04
Pm 2022 05 styling issue

phillimorland:


phillimorland:
@fuzzylogic2000 I am wondering if we should also look at uninstalling bootstrap? it seems like we only use margins and paddings and we can set those ourselves in a utilities file

fuzzylogic2000:
> @fuzzylogic2000 I am wondering if we should also look at uninstalling bootstrap? it seems like we only use margins and paddings and we can set those ourselves in a utilities file

@phillimorland Yes, I think that is a good idea! Then it's also easier to find the styles that should be different and be sure, there is no unintended styling for headlines or links or whatever,

fuzzylogic2000 (approved):
Looks awesome! 
(But I did not see the paralaxing, I guess you'll have to tell me how to do the blocks, so that I can see it.)

phillimorland:
@fuzzylogic2000 parralexing only applies to the hero image, as you scroll you see more of the image but it doesn't continue to the other blocks, this you need a big hero image for it, I took the one from nextcloud

phillimorland:
@fuzzylogic2000 I will remove bootstrap in a separate PR

PR #433, opened 2022-05-05
Pm 2022 05 deps upate

phillimorland:


digitalstrategie/templates/includes/header_language_switch.html
```diff
@@ -2,30 +2,44 @@
...
+      </form>
   </li>
-{% endfor %}
+  {% endif %}
+  {% endfor %}
+</ul>
+
+<h3 class="heading">{% trans 'Accessibility' %}</h3>
+<ul>
+    {% for lang_code, lang_name in languages %}
```
phillimorland, line 28: couldn't think of another way to do this while keeping the required html format? double loop is not cool though i know, any ideas? it does need to be in separate ul thats the issue

fuzzylogic2000: I found a way! Will add it as an extra commit, so we have to make it nice and amend that to the first commit here on Monday.

fuzzylogic2000 (commented):
Supercool! 
Just the green in the green blocks does not look right. Will find out about that.
Screenshot from 2022-05-06 08-24-37

digitalstrategie/templates/includes/header_language_switch.html
```diff
@@ -25,21 +25,17 @@ <h3 class="heading">{% trans 'Languages' %}</h3>
 
 <h3 class="heading">{% trans 'Accessibility' %}</h3>
 <ul>
-    {% for lang_code, lang_name in languages %}
-    {% get_language_info for lang_code as language %}
-    {% if language.code == 'de-ls' %}
+    {% get_language_info for 'de-ls' as language %}
```
phillimorland, line 28: super awesome! :)

phillimorland (commented):
looking good with your fixes!

PR #436, opened 2022-05-05
Kl 2022 05 issues

fuzzylogic2000:


digitalstrategie/assets/scss/components/_ds-block.scss
```diff
@@ -101,3 +101,11 @@
 .hero__image-back {
     z-index: -1;
 }
+
+.ds-block__paragraph {
+    h2:nth-child(1),
+    h3:nth-child(1),
+    h4:nth-child(1) {
+        margin-top: 0 !important;
+    }
```
phillimorland, line 110: very nice!

apps/home/templates/apps_home/blocks/paragraph_block.html
```diff
@@ -0,0 +1,7 @@
+{% load block_tags %}
+
+<div class="modul-servicepanel fullwidth ds-block ds-block--body-bg">
```
phillimorland, line 3: service panel and full width we used for elements that needed to have the coloured background over the full width of the page, as we no longer need that, I would now use their `modul-text_bild` just without a bild like they do in their article example page

phillimorland (commented):
very nice! just small adjustment to paragraph template to keep inline with their styling

apps/home/templates/apps_home/blocks/paragraph_block.html
fuzzylogic2000: So, I replace the classes that are there now with `modul-text_bild ds-block ds-block--body-bg`, right? Will do!

fuzzylogic2000:
Ok, I will merge as you kind of approved @phillimorland when I did the changes (which I did) and I need the migrations.

apps/home/templates/apps_home/blocks/paragraph_block.html
```diff
@@ -0,0 +1,7 @@
+{% load block_tags %}
+
+<div class="modul-text_bild ds-block ds-block--body-bg">
+    <div class="servicepanel__main ds-block__paragraph">
+        {{ self.body }}
+    </div>
+</div>
```
phillimorland, line 7: sorry I meant to use their markup, not just class name and sorry didn't add a link, not sure we should because it's public? anyway something like 
```
<section class="modul-text_bild">
    <div class="text textile">{{ self.body }}</div>
</section>
```

fuzzylogic2000: Ah, cool, and sorry for merging! It looked fine in this PR, but not anymore with the headline anyway, so I will do that in that PR. @phillimorland

PR #439, opened 2022-05-09
a11y: add required titles (h1) to all page types

fuzzylogic2000:
- [x] need to properly style the paragraphs on the detail page


khamui:
nice!

PR #445, opened 2022-05-11
#6005: adding helptext to richtextblocks

khamui:


Rineee (changes requested):
The OverviewPage and GruenbuchOverviewPage also have a RichTextField (page_intro), we should also add it there, I guess?

apps/gruenbuch/models.py
```diff
@@ -42,13 +46,16 @@ class GruenbuchOverviewPage(MetadataPageMixin, Page):
 
     page_intro_de = fields.RichTextField(
         blank=True, default="",
-        verbose_name="Grünbuch Overview page introduction")
+        verbose_name="Grünbuch Overview page introduction",
+        help_text=HELPTEXT_RICHTEXT_A11Y)
```
Rineee, line 50: You could also use the string directly from apps_blocks, so we only define it once and only have to change it once in case that is needed... so here apps_blocks.HELPTEXT_RICHTEXT_A11Y and then remove it from above.. that should not change the migrations.

apps/home/models.py
```diff
@@ -157,11 +162,14 @@ class OverviewPage(MetadataPageMixin, Page):
         max_length=120, blank=True)
 
     page_intro_de = fields.RichTextField(
-        blank=True, default="", verbose_name="Overview page introduction")
+        blank=True, default="", verbose_name="Overview page introduction",
+        help_text=HELPTEXT_RICHTEXT_A11Y)
```
Rineee, line 166: Same here..

Rineee:
@khamui Sorry, another tiny thing :grimacing:

Rineee (approved):
Yeah!

PR #446, opened 2022-05-12
add newsletter block and api

goapunk:


goapunk:
For testing you have to fill in the settings (settings -> newsletter) in wagtail admin

apps/contrib/views.py
```diff
@@ -13,3 +21,30 @@ def get_queryset(self):
         query = self.request.GET.get('q')
         res = sb.search(query, Page, fields=get_search_fields())
         return res
+
+
+def newsletter_view(request):
```
goapunk, line 26: I put this here, but maybe it should be in apps.home instead of apps.contrib?

apps/home/templates/apps_home/blocks/newsletter_block.html
```diff
@@ -10,92 +10,33 @@ <h2 class="title">
         </h2>
         {% endif %}
 
-        <form name="form1" id="form1" action="/newsletter-signup/" method="POST" >
-          <label for="id_email" class="form-label">{{ self.email_field_label }}<span class="icon-required" aria-hidden="true">*</span><span class="aural">({% trans 'required' %})</span></label>
-          <input class="form-control" type="text" id="id_email" name="email" required="" aria-required="true">
-
-          <div class="message message--error u-d-none" id="id_email_invalid" role="alert">{% trans 'Please fill in valid email address.' %}</div>
-          <div class="message message--success u-d-none" id="id_email_valid" role="alert">{% trans 'Your subscription to our E-Mail Newsletter was successful.' %}</div>
+        <form name="form1" id="newsletter_form" action="/newsletter-signup/" method="POST" >
```
khamui, line 13: in order to avoid user-agent validation, maybe adding `novalidate` here?
otherwise we have double validation errors

apps/contrib/views.py
```diff
@@ -13,3 +21,30 @@ def get_queryset(self):
...
+            configuration = sib_api_v3_sdk.Configuration()
+            configuration.api_key['api-key'] =\
+                newsletter_settings.sendinblue_api_key
+            api_instance = sib_api_v3_sdk.ContactsApi(
+                sib_api_v3_sdk.ApiClient(configuration))
+            create_doi_contact = sib_api_v3_sdk.CreateDoiContact(
+                email=form.cleaned_data['email'],
+                include_list_ids=[newsletter_settings.sendinblue_list_id],
+                template_id=newsletter_settings.sendinblue_template_id,
+                redirection_url=newsletter_settings.redirect_url)
```
khamui, line 40: was wondering if this redirection is being used? or is it needed for CreateDoiContact?

khamui (commented):
pretty nice! just two tiny questions..

apps/home/templates/apps_home/blocks/newsletter_block.html
khamui: yes, maybe just chrome issue?

fuzzylogic2000 (approved):
Does look good. In the code and on the site. I am fine with merging without moving the method.

apps/home/templates/apps_home/blocks/newsletter_block.html
phillimorland: ok from that I found another bug but good call fixed now :)

PR #471, opened 2022-05-30
Kl 2022 translations

fuzzylogic2000:


locale/de/LC_MESSAGES/django.po
```diff
@@ -126,8 +124,6 @@ msgid "Send"
 msgstr "Senden"
 
 #: apps/gruenbuch/templates/gruenbuch_overview_page.html:34
-#, fuzzy
-#| msgid "Download"
 msgid "Downloads"
 msgstr "Herunterladen"
```
fuzzylogic2000, line 128: @hmkay or @maxliqd While fixing the translations, I noticed that (introduced by the fuzzy translations), we translate "Downloads" with "Herunterladen". And I couldn't think of the right translation to use here. Please help?!

hmkay: Where do I find this on the page, @fuzzylogic2000 ? Also, I would be fine with just saying "Downloads". I feel this is a name that is commonly used in Germany for downloads.

Rineee (approved):
Nice!! So you can merge unless you want to change the translation for Downloads here?

PR #493, opened 2022-08-24
[#6604] add captcha to forms

goapunk:
- [x] add captcha to forms
- [ ] add captcha to newsletter block?

### DIFF
--- a/apps/forms/models.py
+++ b/apps/forms/models.py
@@ -8,11 +8,16 @@ from wagtail.admin.panels import FieldRowPanel
 from wagtail.admin.panels import MultiFieldPanel
 from wagtail.admin.panels import ObjectList
 from wagtail.admin.panels import TabbedInterface
+from wagtail.contrib.forms.forms import FormBuilder
 from wagtail.contrib.forms.models import AbstractEmailForm
 from wagtail.contrib.forms.models import AbstractFormField
 
+from apps.captcha.fields import CaptcheckCaptchaField
 from apps.contrib.translations import TranslatedField
 from apps.settings import helpers
+
+CAPTCHA_HELP = _('If you are having difficulty please contact us'
+                 ' by {}email{}.')
 
 
 class FormField(AbstractFormField):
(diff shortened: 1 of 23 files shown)


apps/forms/models.py
```diff
@@ -21,8 +23,23 @@ class FormField(AbstractFormField):
...
 
+class WagtailCaptchaFormBuilder(FormBuilder):
+    @property
+    def formfields(self):
+        # Add captcha to formfields property
+        fields = super().formfields
+        fields['captcha'] = CaptcheckCaptchaField(
+            label=_('I am not a robot'),
+            help_text=_('If you are having difficulty please contact us, '
+                        'details adjacent'))
```
fuzzylogic2000, line 34: I think the details adjacent is true for a+, but not here?

fuzzylogic2000 (commented):
I did rebase, and it is mainly working! But the captcha is not loaded. Did I mess that up with the rebase?

apps/forms/models.py
goapunk: yes, and we also need to remove that anyway as it doesn't pass the bitv test

goapunk:
> I did rebase, and it is mainly working! But the captcha is not loaded. Did I mess that up with the rebase?

it seems I didn't add it anywhere yet, or maybe forgot to commit it :man_shrugging:

goapunk:
ok, I forgot to include the tag which loads the js, with it it works

apps/forms/models.py
goapunk: todo: need wording here

goapunk:
ok, this should be ready for another review

apps/forms/templates/apps_forms/includes/form_field.html
```diff
@@ -33,6 +33,36 @@
     {% endif %}
 </fieldset>
 <hr>
+{% elif field|is_captcha %}
+<div class="form-group">
```
goapunk, line 37: I copied this whole block from a+, so not super sure if we need all of this

phillimorland (commented):
it uses `Ubuntu,Roboto,Arial,sans-serif;` that font stack, so we should probably change that. Also confirmation page causes error but dont think thats to do with captcha
`TypeError at /en/contact-form/
FormSubmission() got an unexpected keyword argument 'email'`

goapunk:
I fixed the errors and removed the font-family

phillimorland (approved):
looking good, it's in english on the German site, once that's sorted good to go

fuzzylogic2000:
@phillimorland @goapunk This is good to go as the translations are done in weblate now, right?

phillimorland:
@fuzzylogic2000 we don't have js translations on DS so not sure?!

goapunk:
@fuzzylogic2000 @phillimorland will add the js translations in weblate

PR #528, opened 2022-10-31
[#6627] deps wagtail4

phillimorland:
branch off wagtail3 branch, that should be merged first


Rineee (approved):
Yeah cool, works both on my db and empty one!

PR #531, opened 2022-10-31
[#6514] add libs and update webpack

phillimorland:


phillimorland:
@goapunk not great but it's working

goapunk (approved):
looks ok to me. I'm sure new issues will arise with it but let's give it a try

PR #535, opened 2022-11-01
[#6598] gruenbuch/models: extend oveview page to create index page with migra…

phillimorland:
…tion

I don't know if this is the best way to do this?!


apps/gruenbuch/models.py
```diff
@@ -103,6 +103,25 @@ class GruenbuchOverviewPage(MetadataPageMixin, Page):
...
+    subpage_types = ['apps_gruenbuch.GruenbuchDetailPage',
+                     'apps_gruenbuch.GruenbuchIndexPage']
+
+    search_fields = Page.search_fields + [
+        index.SearchField('page_title_de', partial_match=True),
+        index.SearchField('page_title_en', partial_match=True),
+        index.SearchField('page_title_de_ls', partial_match=True),
+        index.SearchField('page_intro_de', partial_match=True),
+        index.SearchField('page_intro_en', partial_match=True),
+        index.SearchField('page_intro_de_ls', partial_match=True),
```
phillimorland, line 115: I assume each page model needs it's own search field even if it extends another model right? @goapunk

apps/gruenbuch/models.py
```diff
@@ -115,6 +116,15 @@ class GruenbuchOverviewPage(MetadataPageMixin, Page):
     ]
 
 
+class GruenbuchIndexPage(GruenbuchOverviewPage):
+    template = 'gruenbuch_overview_page.html'
```
Rineee, line 120: Its the same as parent page, right? So do you need that again?

phillimorland: when I didn't have it it automatically looked for a template called index_page which didn't exist

Rineee (approved):
Yeah cool! Even though this structure is confusing, working all fine!

PR #536, opened 2022-11-02
apps/home: add microsite overview and detail pages and remove bg colo…

Rineee:
…r from overview page

No migrations added.


apps/home/models.py
```diff
@@ -105,7 +105,9 @@ class HomePage(MetadataPageMixin, Page):
                      'apps_home.SimplePage',
                      'apps_gruenbuch.GruenbuchOverviewPage',
                      'apps_forms.ContactFormPage',
-                     'apps_forms.ParticipationFormPage']
+                     'apps_forms.ParticipationFormPage',
+                     'apps_home.MicrositeOverviewPage',
+                     'apps_home.MicrositeDetailPage']
```
phillimorland, line 110: do we want the homepage to have microsite detail pages as direct children?

apps/home/templates/apps_home/overview_page.html
```diff
@@ -3,7 +3,7 @@
 
 {% block content %}
 {% image page.intro_image max-1920x560 as intro_image %}
-<div class="fullwidth ds-block--fullwidth {% if page.background_color %}ds-block--{{ self.background_color }}{% else %}ds-block--body-bg{% endif %}">
+<div class="fullwidth ds-block--fullwidth ds-block--body-bg">
```
phillimorland, line 6: I think `ds-block--fullwidth ds-block--body-bg` can also go now as they were added to ensure colour was full width but might need testing

phillimorland (approved):
nice! just couple questions, if all find then good to have migrations added and be merged

apps/home/models.py
Rineee: I wasnt sure, but as we allow DetailPages as well and the editors will only be able to edit pages that are already there (as they are not allowed to edit homepage), I thought its more flexible this way?

apps/home/templates/apps_home/overview_page.html
Rineee: Cool, will remove!

phillimorland (approved):
nice!

PR #539, opened 2022-11-03
WIP wagtail richtext: missing elements <hr> document link

phillimorland:


apps/home/blocks.py
```diff
@@ -83,7 +83,9 @@ class Meta:
 class ParagraphBlock(blocks.StructBlock):
     body = blocks.RichTextBlock(
         required=True,
-        help_text=HELPTEXT_RICHTEXT_A11Y
+        help_text=HELPTEXT_RICHTEXT_A11Y,
+        features=['bold', 'italic', 'h2', 'h3', 'ol', 'ul', 'hr', 'link',
+                  'document-link', 'image']
```
phillimorland, line 88: this comes from 
https://docs.wagtail.org/en/stable/reference/streamfield/blocks.html#wagtail.blocks.RichTextBlock
links to following even though it is referencing RichTextField not Block
https://docs.wagtail.org/en/stable/advanced_topics/customisation/page_editing_interface.html#rich-text-features

apps/home/blocks.py
```diff
@@ -80,9 +81,10 @@ class Meta:
         icon = 'image'
 
 
+# RichTextField required to ensure same richtext options throughout site
+# editing fetaures of RichTextBlock broken in Wagtail4.1
 class ParagraphBlock(blocks.StructBlock):
-    body = blocks.RichTextBlock(
-        required=True,
+    body = fields.RichTextField(
```
phillimorland: found that the RichTextField does use the default config but does not work when added this way

apps/home/blocks.py
```diff
@@ -81,11 +80,11 @@ class Meta:
...
 
-# RichTextField required to ensure same richtext options throughout site
-# editing fetaures of RichTextBlock broken in Wagtail4.1
 class ParagraphBlock(blocks.StructBlock):
-    body = fields.RichTextField(
-        help_text=HELPTEXT_RICHTEXT_A11Y
+    body = blocks.RichTextBlock(
+        required=True,
+        help_text=HELPTEXT_RICHTEXT_A11Y,
+        editor='default'
```
phillimorland: also did not work but no error

phillimorland:
issue is opened https://github.com/wagtail/wagtail/issues/9281

PR #540, opened 2022-11-03
assets/ds_block: fix top margin for coloured blocks on pages with hero

phillimorland:
when fixing spacing didn't check with all block types after header, this fixes if it's anon coloured block after the header


fuzzylogic2000 (approved):
Nice!

PR #544, opened 2022-11-03
urls: add language prefix to urls

Rineee:


digitalstrategie/urls.py
```diff
@@ -17,23 +18,27 @@
     path('admin/', include(wagtailadmin_urls)),
     path('documents/', include(wagtaildocs_urls)),
     path('i18n/', include('django.conf.urls.i18n')),
-    re_path(r'^sitemap\.xml$', wagtail_sitemap),
     re_path(r'^robots\.txt$', TemplateView.as_view(
         template_name='robots.txt',
         content_type="text/plain"), name="robots_file"),
-    path('search/', SearchResultsView.as_view(), name="search"),
     path('newsletter-signup/', newsletter_view, name="newsletter-signup"),
```
Rineee, line 24: I somehow could not make it work for the newsletter signup, but also could not test properly.. do you think this is important to also have the language there, so I should try again @fuzzylogic2000?

digitalstrategie/urls.py
```diff
@@ -17,23 +18,27 @@
...
 
     # For anything not caught by a more specific rule above, hand over to
     # Wagtail's page serving mechanism. This should be the last pattern in
     # the list:
-    path('', include(wagtail_urls)),
+    # path('', include(wagtail_urls)),
 
     # Alternatively, if you want Wagtail pages to be served from a subpath
     # of your site, rather than the site root:
     #    url(r'^pages/', include(wagtail_urls)),
```
fuzzylogic2000, line 33: Can we either move or entirely remove this?

digitalstrategie/urls.py
fuzzylogic2000: Hmm, we are not giving the language to the newsletter provider, right? That was in kosmo...
So, I don't think, we need the language here, but maybe we should talk to @maxliqd again.

fuzzylogic2000 (changes requested):
Yes, it does add the language path, but the language switch does not work anymore. And it looks off, for some reason, the list has the dots now.
Screenshot from 2022-11-04 09-54-15

Rineee:
> Yes, it does add the language path, but the language switch does not work anymore. And it looks off, for some reason, the list has the dots now. Screenshot from 2022-11-04 09-54-15

Yeah, the dots were there before, but looks good on dev now, so will rebase..

And the switching worked for me? Will check again!

digitalstrategie/urls.py
Rineee: I guess? I wasnt sure where this is coming from, so just commented out.. so should I remove it?

fuzzylogic2000 (approved):
Works in firefox, but not in chromium? let's try on dev!

PR #551, opened 2022-11-07
ds-block: add overwrite for bg coloured blocks w images to ensure eve…

### LINKED ISSUE #388: #5649 Create Content Block with image and colourful background
**URL:** https://digitalstrategie-dev.liqd.net/
**important screensize:** *Mobile*
**device & browser:** *Safari Version 15.3 (17612.4.9.1.8)*
**Comment/Question:** *Can we use the same spacing from the sides on top of the image? I think it would look more balanced.
Thanks!
imageandcolor
*

phillimorland:
…n margins fixes #388

### DIFF
--- a/digitalstrategie/assets/scss/components/_ds-block.scss
+++ b/digitalstrategie/assets/scss/components/_ds-block.scss
@@ -33,6 +33,18 @@
         .background-image + .content-container *:not(.button) {
             color: $black;
         }
+
+        @media print {
+            background-color: none !important;
+        }
+
+        @media screen and (max-width: 37.4375rem) {
+            /* stylelint-disable */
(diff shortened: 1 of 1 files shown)


digitalstrategie/assets/scss/components/_ds-block.scss
```diff
@@ -33,6 +33,18 @@
         .background-image + .content-container *:not(.button) {
             color: $black;
         }
+
+        @media print {
+            background-color: none !important;
```
fuzzylogic2000, line 38: I know, nothing to do with this PR, but should we also set the text colour to black to make sure that the green blocks are also printed fine?

fuzzylogic2000 (approved):
Very nice!

Issue #560, opened 2022-11-09
6603 Video Block, two blocks on one page, first block leads to reaction of second.


maxliqd:
**Essential Information**
**URL:**https://gd-dev.liqd.net/de/test-video-block/
**device & browser:**Mac, chrome
**operating system (incl. version):**
**behaviour:** Two video blocks on one page. When I press the play on the first block (youtube) the second one (vimeo) gets glitchy, but does not start playing.
**expected behaviour:** The right one starts, the other one doesn't do anything

**Additional information if relevant**
**user role:**
**important screensize:**
**Comment/Question:**

**Screenshot if design related**:
Since issues are public, make sure to not display personal or secret data.

![image](https://user-images.githubusercontent.com/94360651/200828692-2028636d-8ad3-48ed-8abb-77f4c027e4fb.png)

maxliqd:
I can activate the second video. If I press accept and play on the first video, it adds a glitchy image below the second video  block:

![image](https://user-images.githubusercontent.com/94360651/200830795-abbecd80-534c-4121-86d2-7eefc5b75df2.png)

maxliqd:
I added a third video here: https://gd-dev.liqd.net/de/test-video-block-overview-page/
The last video block is the one that reacts.

PR #561, opened 2022-11-09
video_block: add identifyer to function to ensure correct behaviour f…

phillimorland:
…ixes #560

again hopefully fixes it but checking isn't fully working


fuzzylogic2000 (approved):
Cool!

PR #629, opened 2023-04-06
apps/contrib: make sure lower footer is only loaded if response ok

fuzzylogic2000:
I think this is an improvement, but couldn't properly try.


philli-m (commented):
when testing the lower footer is not shown, it is shown on main however, maybe the timeout is too quick?

fuzzylogic2000:
> when testing the lower footer is not shown, it is shown on main however, maybe the timeout is too quick?

It did work for me locally, but we can also use 2 or 5 seconds? Does that work locally for you? @philli-m

philli-m:
@fuzzylogic2000 2s works

philli-m (approved):
great!

PR #630, opened 2023-04-13
Jd 2023 04 improve footer handling

goapunk:


Rineee (commented):
Cool, looking good to me!
So now we are caching the footer and writing it to a file and update both of that regularly, right? Seems like a good plan :) Lets try!

apps/contrib/templatetags/contrib_tags.py
```diff
@@ -1,18 +1,28 @@
...
-            return ''
-        return mark_safe(response.text)
-    else:
-        return ''
+        footer = cache.get('footer')
+        if footer:
+            return footer
+        filepath = settings.MEDIA_ROOT + "/landesfooter.inc"
+        if not os.path.isfile(filepath):
+            management.call_command('get_footer')
```
Rineee, line 23: But what happens if it timeouts here? Maybe we could write '' to the filepath then, would that work?

goapunk: I added error handling so it returns '' if the management command failed

Rineee: Cool, thanks!

Rineee (approved):

PR #631, opened 2023-04-13
requirements: don't use psycopg2-binary on prod

goapunk:


philli-m (approved):
coolio! just clean up conflicts

PR #648, opened 2023-04-17
apps/home/measures detail page: adding measures detail page and templ…

khamui:
…ate (draft)


philli-m:
@khamui just a quick note on structure, I think this should be a separate app, like the gruenbuch app, what do you think? just because it doesn't really belong to home, it will be more like the academy app in liqd site

khamui:
> @khamui just a quick note on structure, I think this should be a separate app, like the gruenbuch app, what do you think? just because it doesn't really belong to home, it will be more like the academy app in liqd site
> 
> NOTE: I added the app here #649 maybe you can review, delete your migrations and rebase on that one? we can also talk about it tomorrow I'll work off mine to do the tiles

@philli-m good point, yes that makes sense :)

apps/measures/models.py
```diff
@@ -125,3 +149,191 @@ class MeasuresOverviewPage(MetadataPageMixin, Page):
...
+        'responsible_orgs_de_ls',
+    )
+
+    involved_orgs = TranslatedField(
+        'involved_orgs_de',
+        'involved_orgs_en',
+        'involved_orgs_de_ls',
+    )
+
+    date = TranslatedField(
```
philli-m, line 244: 'date' is a bit odd here, the german translates to maybe more something like 'duration' or something? 
As Max gave them to us in german maybe we should check with him what the form fields should be called in en?

apps/measures/models.py
```diff
@@ -125,3 +149,191 @@ class MeasuresOverviewPage(MetadataPageMixin, Page):
...
+        'page_title_de_ls',
+    )
+
+    responsible_orgs = TranslatedField(
+        'responsible_orgs_de',
+        'responsible_orgs_en',
+        'responsible_orgs_de_ls',
+    )
+
+    involved_orgs = TranslatedField(
```
philli-m, line 238: maybe same for below, also as these are seen by editors (who at somepoint may not be max) we can use full length words here

apps/measures/templates/apps_measures/measures_detail_page.html
```diff
@@ -0,0 +1,97 @@
...
+    {{ page.contact_name }}<br />
+    {% if page.contact_email %}
+        {% trans 'Email' %}: {{ page.contact_email }}<br />
+    {% endif %}
+    {% if page.contact_phone %}
+        {% trans 'Phone number' %}: {{ page.contact_phone }}<br />
+    {% endif %}
+    {% if page.contact_website %}
+        {% trans 'Website' %}: {{ page.contact_website }}
+    {% endif %}
```
philli-m, line 95: I would wrap the contact deets in an `<address>` element.
They should also probably be links as makes it more accessible but guess we check with Marcelo first.

apps/measures/templates/apps_measures/measures_detail_page.html
```diff
@@ -0,0 +1,97 @@
...
+        <img
+            src="{{ text_image.url }}"
+            alt="{{ text_image.alt }}"
+        />
+        {% if text_image.caption %}
+            <p class="image__caption">{{ text_image.caption }}</p>
+        {% endif %}
+    </div>
+    {{ page.body|richtext }}
+    {% if page.body_participation %}
```
philli-m, line 74: each of these richtext elements can go into a `<section>` element, I think they have an example of what classes they use in the style guide

apps/measures/models.py
```diff
@@ -10,6 +12,28 @@
...
+    ('4', _('Lichtenberg')),
+    ('5', _('Marzahn-Hellersdorf')),
+    ('6', _('Mitte')),
+    ('7', _('Neukölln')),
+    ('8', _('Pankow')),
+    ('9', _('Reinickendorf')),
+    ('10', _('Spandau')),
+    ('11', _('Steglitz-Zehlendorf')),
+    ('12', _('Tempelhof-Schöneberg')),
+    ('13', _('Treptow-Köpenick')),
```
philli-m, line 34: just a question, why numbers? is it a database thing?

apps/measures/templates/apps_measures/measures_detail_page.html
```diff
@@ -0,0 +1,97 @@
+{% extends "base.html" %}
+{% load wagtailcore_tags wagtailimages_tags i18n %}
+
+{% block content %}
+<div id="layout-grid__area--maincontent">
+    <h1>{{ page.page_title }}</h1>
```
philli-m, line 6: check styleguide or our code for required classes for headers, think its just 'title' same for all elements on page, img, p included

apps/measures/templates/apps_measures/measures_detail_page.html
```diff
@@ -0,0 +1,97 @@
...
+    {% image page.text_image original as text_image %}
+    <div class="meta-information__image">
+        <img
+            src="{{ text_image.url }}"
+            alt="{{ text_image.alt }}"
+        />
+        {% if text_image.caption %}
+            <p class="image__caption">{{ text_image.caption }}</p>
+        {% endif %}
+    </div>
```
philli-m, line 72: I think this should be in the same section as the description and styleguide has classes for a image in a paragraph

philli-m (commented):
Awesome! just a few notes on some wording structure bits as we have to use styleguide as much as possible but otherwise, it's good to go!

apps/measures/models.py
```diff
@@ -125,3 +149,191 @@ class MeasuresOverviewPage(MetadataPageMixin, Page):
         index.SearchField('page_intro_en', partial_match=True),
         index.SearchField('page_intro_de_ls', partial_match=True),
     ]
+
```
philli-m, line 152: on line 142 you can add this as a page type

apps/measures/models.py
```diff
@@ -125,3 +149,191 @@ class MeasuresOverviewPage(MetadataPageMixin, Page):
...
+    involved_orgs_en = models.CharField(
+        max_length=300, blank=True)
+    involved_orgs_de_ls = models.CharField(
+        max_length=300, blank=True)
+
+    districts = MultiSelectField(
+        blank=True,
+        max_choices=3,
+        choices=BEZIRK_CHOICES
+    )
```
philli-m, line 180: @khamui sorry just experienced this issue, maybe help text for may choice is good idea?

apps/measures/models.py
khamui: no, just random. I think from an example.

apps/measures/models.py
khamui: yes, added. not sure about my wordin, though.

philli-m (approved):
Super cool!

PR #656, opened 2023-04-20
[#7260 #7261] add measures details to tile and to filter 

philli-m:


apps/measures/models.py
```diff
@@ -139,7 +139,27 @@ class MeasuresOverviewPage(MetadataPageMixin, Page):
...
+    def measures_pages(self):
+        measures_pages = MeasuresDetailPage.objects.live().descendant_of(self)
+
+        return measures_pages
+
+    def get_context(self, request):
+        measures_pages = self.measures_pages
+        context = super().get_context(request)
+        context['measures_pages'] = measures_pages
+        return context
```
philli-m, line 162: @Rineee @fuzzylogic2000 I just copied this context from liqd and removed what I thought we didn't need but might be clunky, also I assume we will need additional stuff for the filtering?

fuzzylogic2000 (approved):
Yeah, nice!

PR #659, opened 2023-04-24
fix js include for video block

goapunk:
including the js in the block template will cause it to run multiple times if there are multiple videos

### DIFF
--- a/digitalstrategie/templates/base.html
+++ b/digitalstrategie/templates/base.html
@@ -33,6 +33,7 @@
         <script src="//www.berlin.de/i9f/r1/bundle/shariff.min.js"></script>
         <script src="//www.berlin.de/i9f/v4/js/bo-foot-rebrush.js"></script>
 
+        <script src="{% static 'video.js' %}"></script>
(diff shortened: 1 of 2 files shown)


digitalstrategie/templates/base.html
```diff
@@ -33,6 +33,7 @@
         <script src="//www.berlin.de/i9f/r1/bundle/shariff.min.js"></script>
         <script src="//www.berlin.de/i9f/v4/js/bo-foot-rebrush.js"></script>
 
+        <script type="text/javascript" src="{% static 'video.js' %}"></script>
```
khamui, line 36: think you can omit the type, as it is javascript (according to MDN).
But really nit-picky so up to you, feel free to merge otherwise.

khamui (approved):

digitalstrategie/templates/base.html
goapunk: it's more in line with the js we have right above, so will change

khamui:
nice!

PR #660, opened 2023-04-24
apps/measures/detail-page: make styles work for mobile

khamui:


digitalstrategie/assets/scss/components/_measures.scss
```diff
@@ -10,3 +10,14 @@
...
+
+@media screen and (max-width: 37.4375rem) {
+    .meta-information__row {
+        flex-direction: column;
+        margin: 0;
+    }
+
+    .meta-information__column {
+        margin-bottom: ($spacer * 2);
+    }
```
philli-m, line 22: this spacing looks a bit bigger then on the designs tested on chrome, but I also struggled to get the actual amount from zeplin because of the elements being grouped, what do you think?

philli-m (changes requested):
I actually didn't pick up on it before but `_measurs.scss` would normally be called `_meta.scss` as that's the name of the component being styled in it, makes it easier to navigate when there are lots of components, we haven't been strict about it in this project but as this is purely just for the table I think we can be here.

philli-m (approved):
Nice!

PR #674, opened 2023-04-26
add initial content-security-policy

philli-m:
* use django-csp
* add csp for dev to find issues

### DIFF
--- a/digitalstrategie/settings/base.py
+++ b/digitalstrategie/settings/base.py
@@ -69,6 +69,7 @@ MIDDLEWARE = (
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'django.middleware.locale.LocaleMiddleware',
+    'csp.middleware.CSPMiddleware',
     'wagtail.contrib.redirects.middleware.RedirectMiddleware',
 )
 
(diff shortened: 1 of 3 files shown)


philli-m:
@goapunk sorry i restored but now it says it's mine

goapunk:
I added some missing bits, but the error on dev came because the csp here only applies to local testing, therefore it used the strict defaults. Depends on admin

goapunk:
> Lgtm. But I have a question: why are adding all of these to dev settings?

good question, my thought was that this way we can test them locally and find problems quickly. And for prod we need a slightly different one anyway. I guess we could also put it in base and then overwrite it for prod. Any preference?

philli-m:
> I added some missing bits, but the error on dev came because the csp here only applies to local testing, therefore it used the strict defaults. Depends on admin

@goapunk so it's dependent on https://github.com/liqd/admin/pull/605 then?

goapunk:
> > I added some missing bits, but the error on dev came because the csp here only applies to local testing, therefore it used the strict defaults. Depends on admin
> 
> @goapunk so it's dependent on [liqd/admin#605](https://github.com/liqd/admin/pull/605) then?

yes

Rineee:
> > Lgtm. But I have a question: why are adding all of these to dev settings?
> 
> good question, my thought was that this way we can test them locally and find problems quickly. And for prod we need a slightly different one anyway. I guess we could also put it in base and then overwrite it for prod. Any preference?

Ah ok, thanks for explaining. No, I think it makes sense to leave them in dev then :+1:

PR #688, opened 2023-05-02
[7261] apps/measures/index page: add pagination and filters

fuzzylogic2000:


fuzzylogic2000:
Help! I went a bit crazy with the filters (which are crazy themselves) in the second commit and now I cannot see the wood for the trees (I looked it up, also works in English :rofl: ) @Rineee @goapunk Can you have a look and point at the wood if there is one? Ehm, I mean point out what can be improved. I will spent the meantime with building the crazy urls in the frontend.

apps/measures/models.py
```diff
@@ -197,14 +201,73 @@ def facilitative_administration(self):
...
+                fields_filter_params = fields_filter_params | self._q_filter(
+                    inc, 'inclusive_shaping_of_urban_life'
+                )
+            if fac:
+                fields_filter_params = fields_filter_params | self._q_filter(
+                    fac, 'facilitative_administration'
+                )
+            filter_params = filter_params & fields_filter_params
+
+        measures_pages = measures_pages.filter(Q(filter_params))
```
Rineee, line 254: Do we really need the Q around again?

apps/measures/models.py
```diff
@@ -197,14 +201,73 @@ def facilitative_administration(self):
...
-        measures_pages = MeasuresDetailPage.objects.live().descendant_of(self)
+        measures_pages = MeasuresDetailPage.objects.live()\
+            .descendant_of(self).order_by("title", "-first_published_at")
 
         return measures_pages
 
+    def _q_filter(self, filter_list, field_name):
+        filter_params = Q()
+        for element in filter_list:
+            kwargs = {'{0}__contains'.format(field_name): element}
```
Rineee, line 212: Do we need the 0?

Rineee (approved):
Cool, it works really well! 

Only 2 little questions/suggestions.

And I was trying to write the filtering stuff for the lists a bit shorter, but did not really succeed :) A maybe a little shorter way would be sth like [here](https://stackoverflow.com/a/4824810), but yeah, not a super great improvement

apps/measures/models.py
fuzzylogic2000: You are right, that should be safely Q already now. :rofl:

apps/measures/models.py
fuzzylogic2000: No, with one element, it's also readable without.

Rineee:
> Cool, it works really well!
> 
> Only 2 little questions/suggestions.
> 
> And I was trying to write the filtering stuff for the lists a bit shorter, but did not really succeed :) A maybe a little shorter way would be sth like [here](https://stackoverflow.com/a/4824810), but yeah, not a super great improvement

Couldnt help but trying out a bit :see_no_evil: and came up with this MASSIVE improvement :rofl: 
`    def _q_filter(self, filter_list, field_name):
        kwargs = [('{}__contains'.format(field_name),element) for element in filter_list]
        return Q(*kwargs, _connector=Q.OR)
`
But yeah, for readibility probably yours is better, so I am totally fine with both!

fuzzylogic2000:
> 

I added a docstring now, and already find my code quite hard to read, so @Rineee yours is not much worse IMHO. But `Readability counts.` Ah, don't know. @goapunk please decide!

goapunk:
> I added a docstring now, and already find my code quite hard to read, so @Rineee yours is not much worse IMHO. But `Readability counts.` Ah, don't know. @goapunk please decide!

I'd say lets go for the more readable version :)

fuzzylogic2000:
> > I added a docstring now, and already find my code quite hard to read, so @Rineee yours is not much worse IMHO. But `Readability counts.` Ah, don't know. @goapunk please decide!
> 
> I'd say lets go for the more readable version :)

Alright, thanks! I will merge then. 

@Rineee And I did change your other suggestions. :)

PR #692, opened 2023-05-03
[#7267 #7264] serve nav menu dependent on url and add ls to footer

philli-m:
NOTE: This is does not include the functionality to go back to LS homepage if no LS content on current page when lang switch changed.


apps/snippets/models.py
```diff
@@ -34,9 +34,9 @@ class MenuItem(models.Model):
         blank=True,
         null=True,
         help_text=(
-            'Creates a link to a single wagtail page. '
+            'Header: creates a link to a single wagtail page. '
             'Leave empty if you add subpages. '
-            'In the footers only subpages will be shown.'
+            'Footer: only subpage links will be shown.'
```
philli-m, line 39: I think this is clearer but i'm not sure?!

apps/snippets/models.py
```diff
@@ -96,10 +96,12 @@ class NavigationMenu(ClusterableModel):
         max_length=255,
         null=False,
         blank=False,
-        help_text=_('Call the menu "header" to use as upper menu. '
-                    'Call the menu "footer" to use as'
-                    'the upper light grey footer. The lower '
-                    'footer is loaded externally.')
+        help_text=_('Call the menu "header" for use on de and en pages '
+                   'and header_ls for ls pages as the upper menu. '
```
Rineee, line 100: really nit-picking, but could we also add quotation marks here (also for footer) to be consistent?

Rineee:
I added a commit that adds a template tag to decide if there is content. Although it relies on the naming of the panels, which is not ideal, but I think there is no other way? I added a comment, although a test would probably be better :rofl: Should we document it somewhere else?  

I spoke to Max and he said that if there is any content, the page should be displayed as it is very unlikely that only e.g. the title will be filled. I guess thats true, but maybe a helptext in the easy language tab could also explain the behaviour that the page is displayed if any of the fields are filled?

fuzzylogic2000 (approved):
NIce! Will change the string and merge.

PR #694, opened 2023-05-03
apps/*/models: add syntax for translating all strings in the wagtail …

philli-m:
…admin
as discussed with Max ensuring the admin is translated seemed wise as we already added partial translations


apps/gruenbuch/models.py
```diff
@@ -42,15 +43,15 @@ class GruenbuchOverviewPage(MetadataPageMixin, Page):
 
     page_intro_de = fields.RichTextField(
         blank=True, default="",
-        verbose_name="Page introduction de",
+        verbose_name=_('Page introduction de'),
         help_text=apps_blocks.HELPTEXT_RICHTEXT_A11Y)
```
khamui, line 47: would it make sense to translate the constants too?

khamui (approved):
nice! just a question regarding a11y constant. if not relevant, feel free to merge.

apps/gruenbuch/models.py
philli-m: oupps missed the blocks.py translations, nice one, will add :)

PR #696, opened 2023-05-04
apps/*OverviewPages/templates: check for language_code and in case of…

khamui:
… de-ls hide intro image

### DIFF
--- a/apps/home/templates/apps_home/overview_page.html
+++ b/apps/home/templates/apps_home/overview_page.html
@@ -6,7 +6,7 @@
 <div class="fullwidth">
     <div id="layout-grid">
     <section id="layout-grid__area--herounit">
-        {% if page.intro_image %}
+        {% if page.intro_image and request.LANGUAGE_CODE != 'de-ls' %}
         <div class="image">
             <div class="image__image">
                 <img src="{{ intro_image.url }}" class="w-100" alt="{{ page.intro_image.alt }}"/>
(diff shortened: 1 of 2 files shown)


apps/home/templates/apps_home/overview_page.html
```diff
@@ -6,7 +6,8 @@
 <div class="fullwidth">
     <div id="layout-grid">
     <section id="layout-grid__area--herounit">
-        {% if page.intro_image %}
+        {{ request.LANGUAGE_CODE }}
+        {% if page.intro_image and request.LANGUAGE_CODE != 'de-ls' %}
```
philli-m, line 10: why use a variable and template tag here and just a template tag above, do I miss something?
also the language switch doesn't seem to work on overview anymore, does on gruenbuch though, maybe doing it other was is better?

philli-m (commented):
When testing I had a thought, do we definitely not want this on the microsites overview too?

apps/home/templates/apps_home/overview_page.html
fuzzylogic2000: @khamui What is the status here? Did you resolve it, can I have another look? And do you also want to add it to the microsites or did you already?

khamui: yes @fuzzylogic2000 it is resolved and works as expected.

fuzzylogic2000 (approved):
Works!

PR #709, opened 2023-05-09
apps/measures/overview: add search

fuzzylogic2000:
The cross to remove the search is still missing, will add it as a task

Right, and you need elasricsearch running for that.
Try:
`sudo docker run --name es01-test -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" -e "logger.level=DEBUG" docker.elastic.co/elasticsearch/elasticsearch:7.17.10`
if that tells you, that the container is already in use, delete it: 
`docker rm -f es01-test`
And again the thing from above. When elasticsearch is running, in another window:
`python manage.py update_index`
And now with that you are ready to also `make watch` and try the search!

### DIFF
--- a/apps/contrib/templatetags/contrib_tags.py
+++ b/apps/contrib/templatetags/contrib_tags.py
@@ -41,6 +41,18 @@ def combined_url_parameter(request_query_dict, **kwargs):
 
 
 @register.simple_tag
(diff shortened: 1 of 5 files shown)


philli-m (approved):
looking fab! I just added a line as looks like in the example page we have they added their own styling for the button.

PR #710, opened 2023-05-09
apps/measures/fields of action: making links on detail page lead to f…

khamui:
…iltered overview page

### DIFF
--- a/apps/measures/models.py
+++ b/apps/measures/models.py
@@ -518,9 +518,31 @@ class MeasuresDetailPage(Page):
     ]
 
     @property
-    def get_fields_of_action_display(self):
-        fields_of_action = [self.get_regenerative_management_display(),
-                            self.get_future_opportunities_for_all_display(),
-                            self.get_inclusive_shaping_of_urban_life_display(),
-                            self.get_facilitative_administration_display()]
-        return ', '.join(filter(None, fields_of_action))
+    def get_fields_of_action(self):
+        """Return a dict of selected fields of action and their subsets."""
+        fields_of_action = self.get_selected_foa(
+            self.regenerative_management, 'reg',
+            REGENERATIVE_MANAGEMENT
+        ) + self.get_selected_foa(
+            self.future_opportunities_for_all, 'fut',
+            FUTURE_OPPORTUNITIES_FOR_ALL
+        ) + self.get_selected_foa(
+            self.inclusive_shaping_of_urban_life, 'inc',
+            INCLUSIVE_SHAPING_OF_URBAN_LIFE
+        ) + self.get_selected_foa(
+            self.facilitative_administration, 'fac',
+            FACILITATIVE_ADMINISTRATION
+        )
+        return fields_of_action
+
+    def get_selected_foa(self, selected_foa_ids, subset_id, subset):
+        """Return an object for given fields of action and given subset."""
+        selected_foa = []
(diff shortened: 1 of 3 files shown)


apps/measures/models.py
```diff
@@ -518,9 +518,29 @@ class MeasuresDetailPage(Page):
...
 
     @property
-    def get_fields_of_action_display(self):
-        fields_of_action = [self.get_regenerative_management_display(),
-                            self.get_future_opportunities_for_all_display(),
-                            self.get_inclusive_shaping_of_urban_life_display(),
-                            self.get_facilitative_administration_display()]
-        return ', '.join(filter(None, fields_of_action))
+    def get_fields_of_action(self):
+        fields_of_action = self.get_selected_foa(
```
fuzzylogic2000, line 522: Could you add a docstring sth. like
`Return all selected fields_of_action in the form ....`
or `Return a dictianary of all selected measures and their sets`

apps/measures/models.py
```diff
@@ -518,9 +518,29 @@ class MeasuresDetailPage(Page):
...
+        ) + self.get_selected_foa(
+            self.inclusive_shaping_of_urban_life, 'inc',
+            INCLUSIVE_SHAPING_OF_URBAN_LIFE
+        ) + self.get_selected_foa(
+            self.facilitative_administration, 'fac',
+            FACILITATIVE_ADMINISTRATION
+        )
+        return fields_of_action
+
+    def get_selected_foa(self, selected_foa_ids, kind_id, kind):
```
fuzzylogic2000, line 537: The `kind` did irritate me a bit. Can we rename that to maybe `set` or `category`? Or @philli-m has an idea how to name the groups or sets of fields of action?

apps/measures/models.py
```diff
@@ -518,9 +518,29 @@ class MeasuresDetailPage(Page):
...
+        selected_foa = []
```
fuzzylogic2000, line 538: And a docstring here!

fuzzylogic2000 (commented):
NIce! I just find the naming a bit confusing and some documentation would be good.

apps/measures/models.py
```diff
@@ -518,9 +518,31 @@ class MeasuresDetailPage(Page):
...
 
     @property
-    def get_fields_of_action_display(self):
-        fields_of_action = [self.get_regenerative_management_display(),
-                            self.get_future_opportunities_for_all_display(),
-                            self.get_inclusive_shaping_of_urban_life_display(),
-                            self.get_facilitative_administration_display()]
-        return ', '.join(filter(None, fields_of_action))
+    def get_fields_of_action(self):
+        """Return a dictionary of selected measures and their categories."""
```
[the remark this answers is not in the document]
khamui: @fuzzylogic2000 ahh..true, sorry. Yes it should be `selected fields of action` instead

fuzzylogic2000 (approved):
Did you want to change it to subset or should we merge like this? I am fine eith either.

fuzzylogic2000:
> > Did you want to change it to subset or should we merge like this? I am fine eith either.
> 
> oh..i forgot, will change it!

You didn't push it, right? :stuck_out_tongue_closed_eyes:

khamui:
@fuzzylogic2000 oh true....uppsala

fuzzylogic2000 (approved):
Nice!

PR #717, opened 2023-05-11
add matomo

goapunk:
* move to own entry point
* adjust CSP


apps/contrib/templates/matomo/tracking_code.html
```diff
@@ -0,0 +1,4 @@
+{% load static %}
+<!-- Matomo -->
+<script id="matomo" data-cookieDisabled={{ cookie_disabled }} data-url="{{ url }}" data-id="{{ id }}"async src="{% static 'matomo.js' %}"></script>
```
goapunk, line 3: feels a bit hacky to put the data attributes on the script tag but worked for me in chrome and firefox. Should be tested in safari as well

fuzzylogic2000: @philli-m or @khamui Can you test this also in safari, please?

PR #718, opened 2023-05-11
[7360] apps//contrib/mixins//home/models//measures/models//gruenbuch/models:…

khamui:
… new mixin TranslatedMetadataPageMixin to replace MetadataPageMixin


apps/contrib/mixins.py
```diff
@@ -0,0 +1,50 @@
+from django.db import models
+from django.utils import translation
+from django.utils.translation import gettext_lazy as _
+from wagtail.admin.panels import FieldPanel
+from wagtail.admin.panels import MultiFieldPanel
+from wagtailmetadata.models import MetadataPageMixin
+
+
+class TranslatedMetadataPageMixin(MetadataPageMixin, models.Model):
```
Rineee, line 9: Since we are basically overwriting everything from MetadataPageMixin here, I would rather inherit from WagtailImageMetadataMixin and also add the missing stuff here (i.e. the image and the methods).

apps/contrib/mixins.py
```diff
@@ -0,0 +1,50 @@
...
+        ], _('Common page configuration')),
+    ]
+
+    def get_meta_title(self):
+        if (translation.get_language() == 'en'):
+            return self.seo_title_en
+        return self.seo_title_de
+
+    def get_meta_description(self):
+        if (translation.get_language() == 'en'):
```
Rineee, line 45: dont need the brackets here :)

Rineee (commented):
Cool, it works, I only have one suggestion.

And another question: so now the old meta tags are lost, we could also write a custom migration to migrate them to the now German fields. But did you maybe already discuss this with @fuzzylogic2000 or/and @maxliqd ?

fuzzylogic2000:
> Cool, it works, I only have one suggestion.
> 
> And another question: so now the old meta tags are lost, we could also write a custom migration to migrate them to the now German fields. But did you maybe already discuss this with @fuzzylogic2000 or/and @maxliqd ?

No, we didn't discuss that, but that's a good idea! Do you want to try @khamui or should @Rineee or me take over?

khamui:
> > Cool, it works, I only have one suggestion.
> > And another question: so now the old meta tags are lost, we could also write a custom migration to migrate them to the now German fields. But did you maybe already discuss this with @fuzzylogic2000 or/and @maxliqd ?
> 
> Oh, and I just realized that it does not work for the 'Leichte Sprache'. You would also need to return the German meta information, when the chosen language is 'Leichte Sprache'. @khamui

it seems, that it works for me. Selecting de-ls sets meta info to German for me. Or maybe I am not following or miss something @Rineee ?

khamui:
> 

i will give it a time boxed try. so either today or I will ask you to take over :) thanks

khamui:
@Rineee @fuzzylogic2000 okay I changed to TranslatedField, but I do not manage to make the custom migration. Hope it is okay to pass to one of you.

fuzzylogic2000:
@Rineee Now it also has migrations, could you check again?

Rineee (approved):
Yeah!

PR #720, opened 2023-05-15
[7362] add sentences to show search, filters and result count and some measure overview related fixes

fuzzylogic2000:
Not sure, I succeeded with the second commit and made it really more readable. Well, I tried.

### DIFF
--- a/digitalstrategie/templates/search_results.html
+++ b/digitalstrategie/templates/search_results.html
@@ -1,33 +1,40 @@
 {% extends "base.html" %}
-{% load wagtailcore_tags i18n %}
+{% load wagtailcore_tags i18n contrib_tags %}
 
 {% block meta_tag %}
-    <title>{% trans 'Gemeinsam Digital: Berlin - full-text search' %}</title>
+    <title>{% translate 'Gemeinsam Digital: Berlin - full-text search' %}</title>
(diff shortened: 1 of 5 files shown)


digitalstrategie/templates/search_results.html
```diff
@@ -1,33 +1,40 @@
...
                             <i class="fas fa-search lens" aria-hidden="true"></i>
                             <input
                               class="form-control search"
                               value="{% if request.GET.q %}{{ request.GET.q }}{% endif %}"
-                              placeholder="{% trans 'Search term' %}"
+                              placeholder="{% translate 'Search term' %}"
                               name="q"
-                            >
+                            />
+                            <button class="button button--clean button--clear-input" type="button" title="{% translate 'Empty field' %}">
```
goapunk, line 28: More of a thought: do we want to always show this? I somehow would expect it to only be there if you actually searched for something.

goapunk (approved):
very nice

digitalstrategie/templates/search_results.html
fuzzylogic2000: @goapunk Yeah, you are right. But the example also always has it. So, let's see if someone complains. :)

PR #725, opened 2023-05-16
readme: update with local elastic search instructions

fuzzylogic2000:


README.md
```diff
@@ -1,21 +1,38 @@
...
+```
+$ make watch
+```
+
+Start local server with elastic search:
+
+Open two terminal windows.
+1. Start elastic search in the first one:
+```
+$ sudo docker run --name es01-test -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" -e "logger.level=DEBUG" docker.elastic.co/elasticsearch/elasticsearch:7.17.10
```
goapunk, line 20: if we remove the `--name` argument we don't need the part with the deletion below

goapunk (changes requested):
looks good to me, just a small comment about the name

README.md
goapunk: ah, sorry, another thing which is probably nice: replace the `-p` wiith `-p 127.0.0.1:9200:9200 -p 127.0.0.1:9300:9300`

goapunk (approved):
cool!

PR #730, opened 2023-05-17
Pm 2023 5 a11y issue

philli-m:


apps/home/templates/apps_home/home_page.html
```diff
@@ -9,7 +9,7 @@
     {% image page.ls_header_image original as header_image %}
 {% endif %}
 <section id="layout-grid__area--herounit" class="fullwidth">
-  <div class="hero__parallax" style="background-image: url('{{ header_image.url }}')" alt="{{ header_image.alt }}" loading="lazy">
```
goapunk, line 12: out of curiosity: why remove the lazy loading?

philli-m: it's an img tag attribute, not div attribute so triggers error

PR #748, opened 2023-05-23
[7361] metatags for detail pages

fuzzylogic2000:
Making the meta tags not rqeuired was a bit late after the other migrations, but I didn't want to redo everything. And it would have been two migrations per app anyway (one for updating and one for adding it to the detail pages of every app).

This might also be a fix for #749 - to be checked after merge.


apps/contrib/templatetags/custom_metadata_tags.py
```diff
@@ -18,11 +17,9 @@ def custom_meta_tags(context, model=None):
...
         model = context.get('self', None)
-        if not hasattr(model, 'get_meta_title') or (
-                not model.meta_title and
+        while not hasattr(model, 'get_meta_title') or (
+                not model.get_meta_title() and
                 not model.get_meta_description()):
-            site = Site.find_for_request(request)
-            if site:
-                model = site.root_page.specific
+            model = model.specific.get_parent().specific
```
Rineee, line 23: I tested an overview page without meta tags first and get an error here: `'NoneType' object has no attribute 'specific'`

Rineee (changes requested):

apps/contrib/templatetags/custom_metadata_tags.py
fuzzylogic2000: Uh, how weird, the overview should have the HomePage as parent, shouldn't it? Without the tags, I only tested the gurenbuch overview thiugh and that was fine. But I'll try all options, we have then!

fuzzylogic2000: Oh right, or the overview itself didn't have the specific? Something weird. Good catch, though!

fuzzylogic2000: I couldn't recreate that. In all versions: overview pages without child pages, with child pages without metatags and with child pages with meta tags. Could we have a look at your page on Tuesday?

(But I found another problem, I fixed. :))

philli-m (approved):
Nice, working well, while testing noticed that we use meta tag instead of title but will make an issue and discuss there.

apps/contrib/templatetags/custom_metadata_tags.py
fuzzylogic2000: @Rineee This was a problem when the root page didn't have meta tags. It should work in any case now. :crossed_fingers:

Rineee: Ok, so now I only get that error on the homepage. So I do need the meta tags for the homepage, do I @fuzzylogic2000 ?

fuzzylogic2000: Hmm, no, I will also catch that!

fuzzylogic2000: Oh, I still had it the wrong way round. Now it's fine!!!!1!!!

fuzzylogic2000: @Rineee Can you please do another round of reviewing?

PR #767, opened 2023-05-31
footer: add parameter to exclude js from the include file

goapunk:
see styleguide wiki for more info on the parameters

fixes broken footer on small screens


philli-m (approved):
Nice!

PR #768, opened 2023-05-31
Ks 05 2023 fix language switch

Rineee:


digitalstrategie/templates/includes/header_language_switch.html
```diff
@@ -8,16 +8,9 @@ <h3 class="heading">{% trans 'Languages' %}</h3>
...
-          {% csrf_token %}
-          <input name="next" type="hidden" value="{{ redirect_to }}" />
-          <input name="language" type="hidden" value="{{ lang_code }}">
-          <span>
-            <button role="button" type="submit" aria-label="{% trans 'select' %}{{ lang_name }}">
-                {{ language.name_local }}
-            </button>
-          </span>
-      </form>
+        <button>
```
Rineee, line 11: this is probably not the right way, but it made it look the same again :)

philli-m: the `header-menu__btn`class we created because the styleguide doesn't define styling for forms in the menu so we had to, as we are now using a link like it suggests we no longer nee to apply any styling, it will work by using their structure of:
```
  {% if language.code != 'de-ls' %}
  <li>
    <a href="{% translate_url lang_code %}" aria-label="{% trans 'select' %}">{{ language.name_local }}</a>
  </li>
  {% endif %}
  ```

philli-m (approved):
Nice!

PR #778, opened 2023-06-05
package.json: override cosmiconfig as workaround for segfault, should…

goapunk:
… be removed later


philli-m (approved):
Looking good and works
@goapunk why does it remove `Segmentation fault: 11` error?

PR #787, opened 2023-06-07
[7361] make metatags go through parents until metatags are found

fuzzylogic2000:
redoes https://github.com/liqd/digitalstrategie/pull/771/files
But checks if the model does not exist which was the problem on the server.


philli-m (approved):
working like a charm, lets check on dev :D

PR #793, opened 2023-06-08
fix various translations

goapunk:
warning: contains migrations


apps/measures/templates/apps_measures/includes/measures_list_item.html
```diff
@@ -25,8 +25,8 @@ <h3 class="title">{{ content.page_title|truncatechars:102 }}</h3>
             </p>
 
             <a href="{% pageurl content %}" class="more" data-mainlink="true" title="This is a single Teaser">
-                {% trans 'More' %}
-                <span class="aural">{% trans 'to: ' %} {{ content.translated_title }}</span>
```
goapunk, line 29: translated_title always seemed to be empty? not sure if that was the correct field so changed it to page_title

goapunk: and for me the SR reads "more to <page title>", somehow "more about <page title>" sounds better to me. What do you think?

Rineee (approved):
Lgtm! Although not sure about the screenreader thing, maybe @philli-m could have a quick look at that?

apps/measures/templates/apps_measures/includes/measures_list_item.html
philli-m: It's just what the styleguide does to more links but happy with the change

philli-m (approved):
cool

PR #821, opened 2023-06-26
measures: search before filtering

### LINKED ISSUE #819: filter and search measures overview results
**Essential Information** If you combine search, project status, district and fields of action, you get contradictory results.
**URL:**https://gd-stage.liqd.net/en/massnahmen-der-strategie/?search=digital&status=&district=1&inc=1&page=1#measures
**device & browser:** chrome, mac os
**operating system (incl. version):**
**behaviour:** wrong number (2) tiles shown, pagination for two pages, text indicates seven results. These seven. results seem to relate to the search, but they are not shown. The pagination seems to depend on the seven results. Filter and search results are not combined: Tiles seem to relate to filter.
**expected behaviour:**  Number in the text should correspond to the number of tiles, pagination also. User should get combined results of search and filter. E. g. all "citywide" (filte district) and "Participation" (filter fields of action) measures that mention the word "digital". Pagination should start for more than six tiles. 

**Additional information if relevant**
**user role:**
**important screensize:**
**Comment/Question:**
Related stories: 
search https://taiga.liqd.net/project/liqd-liquid-software/us/7262
filter https://taiga.liqd.net/project/liqd-liquid-software/us/7261
search and filter results: https://taiga.liqd.net/project/liqd-liquid-software/us/7362

**Screenshot if design related**:
Since issues are public, make sure to not display personal or secret data.
image

goapunk:
Use the search results to filter the measures as previously it would return the wrong number of results (probably a bug in the wagtail search backend).

fixes #819

### DIFF
--- a/apps/measures/models.py
+++ b/apps/measures/models.py
@@ -266,6 +266,18 @@ class MeasuresOverviewPage(TranslatedMetadataPageMixin, Page):
     def get_context(self, request):
         filtered_measures = self.measures_pages
 
+        # search
+        search = request.GET.get('search')
+        search_string = ''
+        if search:
+            searched_measures = filtered_measures.search(search)
+            search_string = _('Search for "{search_term}".').format(
+                search_term=search
+            )
+            filtered_measures = filtered_measures.filter(
+                id__in=[m.id for m in searched_measures]
+            )
+
         # filters for status, district and fields of action (four sets)
         status = request.GET.get('status')
         dis = request.GET.getlist('district')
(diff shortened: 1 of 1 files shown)


apps/measures/models.py
```diff
@@ -266,6 +266,18 @@ def get_foa_filter_info(self, reg, fut, inc, fac):
...
+        search_string = ''
+        if search:
+            searched_measures = filtered_measures.search(search)
+            search_string = _('Search for "{search_term}".').format(
+                search_term=search
+            )
+            filtered_measures = filtered_measures.filter(
+                id__in=[m.id for m in searched_measures]
+            )
+
```
Rineee, line 280: Is that sth else than just filtered_measures = searched_measures?

goapunk: searched_measures are of type ElasticSearchResults whereas filtered_measures is a query set. We can't `filter()` on the SearchResults

PR #824, opened 2023-06-27
settings: use https for csp

goapunk:


philli-m (approved):
looks good I guess we just test on dev? not sure on the youtube shares?!

goapunk:
@philli-m I think youtube should be fine. If they add http links they will have to fix it anyway

philli-m:
cool

PR #872, opened 2023-10-04
makefile: add media folder on make install to ensure app can find foo…

philli-m:
…ter on first make watch


goapunk (changes requested):
needs update to be same as in mb now

philli-m:
@goapunk a little more context?

goapunk (approved):
LGTM

PR #880, opened 2023-10-25
[7759] wagtail update

goapunk:


apps/contrib/views.py
```diff
@@ -11,5 +11,5 @@ class SearchResultsView(ListView):
     def get_queryset(self):
         sb = get_search_backend()
         query = self.request.GET.get('q')
-        res = sb.search(query, Page, fields=get_search_fields())
+        res = sb.autocomplete(query, Page, fields=get_search_fields())
```
goapunk, line 14: The search changes will need some manual testing, I haven't done that yet

goapunk:
ok, the search needs more fixing

apps/contrib/views.py
m4ra: I did a minimal testing locally LGTM

m4ra (approved):

PR #884, opened 2023-10-25
[7774] Make wagtail blocks available on different page types

hom3mad3:


philli-m (approved):
looking good, I would just add the migration renames to the commits where files added and it's good to go

philli-m:
@hom3mad3 also another fun thing about working with wagtail and it's constant need for migrations is that it's always a good idea to se if anything has been merged with migrations since you pulled, as the wagtail update just go merged you will need to delete your migrations you made for this pr and pull the most recent main and install and remake your migrations as migrations on main must always be sequential

philli-m (approved):
great! :)

PR #888, opened 2023-10-26
[issues ] front end release issues

philli-m:
to-do 

- [ ] 1. create admin PR if CSP works on dev


goapunk (changes requested):
LGTM but missing changelog

goapunk (approved):
LGTM

PR #912, opened 2023-10-31
ci translation check

goapunk:


philli-m (approved):
very nice! should we also add this to other projects?

goapunk:
> very nice! should we also add this to other projects?

yup

PR #913, opened 2023-10-31
apps/home/blocks: update teasers to require colour choice fixes #905

### LINKED ISSUE #905: 7774 teaser image background colours not selectd, black
**Essential Information**
**URL:** https://gd-dev.liqd.net/en/wagtail-test-october-23-overview-page/
**device & browser:** mac chrome
**operating system (incl. version):**
**behaviour:** When I don't choose a colour for the background, the background is black
**expected behaviour:**
Two options, easier one can be chosen:
1. Colour should be white, to be in sync with help text
2. Make choosing colour mandatory. Preselect pink, if preselect is necessary.
3. Adapt help text: Choose yellow, pink or purple to ensure CI conformity.

**Additional information if relevant**
**user role:**
**important screensize:**
**Comment/Question:**
**Screenshot if design related**:
Since issues are public, make sure to not display personal or secret data.

philli-m:
after discussion with @maxliqd we also decided to save confusion updating the block labels would be advantageous too, as migrations were needed for the fix I added them to same PR


philli-m:
migrations don't conflict with django update so pushed, will leave rest to you guys

apps/home/blocks.py
```diff
@@ -138,16 +138,17 @@ class TeaserBlockCentered(blocks.StructBlock):
         required=False
     )
     background_color = ColorChoiceBlock(
-        help_text=_('Not choosing a colour will result in a '
-                    'block with a white background.'),
+        help_text=_('Choose yellow, pink or purple to ensure CI conformity.'),
         required=False
```
goapunk, line 142: the help text reads a bit like this should be required?

philli-m: nice catch!

goapunk (approved):
LGTM

PR #928, opened 2023-11-07
apps/contrib: fix elasticsearch showing results from the untranslated…

goapunk:
… title field

currently searching for the english title when on the german or simple german page would still show english results. This should help

### DIFF
--- a/apps/contrib/elasticsearch.py
+++ b/apps/contrib/elasticsearch.py
@@ -4,7 +4,9 @@ from wagtail.search.backends.elasticsearch8 import Elasticsearch8SearchResults
 
 # Code below taken and modified from wagtails elasticsearch backend
 class ElasticsearchResults(Elasticsearch8SearchResults):
+
     def _get_es_body(self, for_count=False):
+        """Override super method to add the highlight config."""
         body = {
             'query': self.query_compiler.get_query()
         }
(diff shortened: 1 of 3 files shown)


apps/contrib/elasticsearch.py
```diff
@@ -25,20 +25,26 @@ def _get_es_body(self, for_count=False):
...
             hs.pop('content_type')
+            # ignore hits from the _all_text field as it doesn't work with
+            # our translations
+            hs.pop('_all_text_boost_2_0')
             for highlight in hs.values():
                 # we only need one highlight
                 if hit['fields']['pk'][0] not in highlights:
                     highlights[str(hit['fields']['pk'][0])] = highlight[0]
+                    # only use search results which have a valid hit
+                    pks.append(hit['fields']['pk'])
```
m4ra, line 43: hard to imagine the data structure without an example and not inline documentation, but looking at the top, isn't here missing a `[0]`?

goapunk: good catch :facepalm:  there's also another mistake. Will push a fix

goapunk: I also added some general comments, but most of the code comes directly from wagtail anyway

m4ra (approved):

PR #949, opened 2023-12-13
templates/home_page: add anchor links for blocks according to WCAG ta…

philli-m:
…ble of contents

tried to mimic how [WCAG](https://www.w3.org/TR/WCAG20/#contents) does their anchor links however could not pass the blocks their id and the stream block does not know the block type so had to use generic link name and have link outside semantic block code, not idea but as blocktype use different structure seems simplest way this should be added to other pages. I think we also need to add help text for editors to identify the block id's in the dashbaord, made sample text but it's a bit long?!

"When you want to add an anchor tag to a section, follow these steps:

    Navigate to the block in the editing interface.
    Click the chain link icon next to the block.
    In the page URL bar, you will see a collection of numbers and letters after the '#' symbol. Copy everything after '#'; this is your anchor link for that section.

Use this copied value as the anchor link when needed."


apps/home/templates/apps_home/home_page.html
```diff
@@ -1,5 +1,5 @@
 {% extends "base.html" %}
-{% load wagtailcore_tags wagtailimages_tags %}
+{% load wagtailcore_tags wagtailimages_tags block_tags %}
```
goapunk, line 2: `block_tags` can be removed

goapunk (approved):
LGTM just the removal of the unused template tag

goapunk:
Maybe a short helptext:
```
To get the anchor tag for this block: click the chain link symbol at the top left of the block and copy the id from the url (example: block-1e495692-5fad-4133-86c5-0c42e19b4ad2-section)
```
?

goapunk (approved):
LGTM

PR #950, opened 2023-12-14
Jd 2023 12 fix image richtext

goapunk:


rittermo:
@sblindow Could you review/test this change, as soon as you have time? I promised them that it would get released by the end of this week. Please let Vincent Ritterspach know, as soon as it is released. I’ll send you his email via Mattermost

PR #951, opened 2023-12-18
templates/home_page: fix paralax scrolling for desktop

philli-m:


apps/home/templates/apps_home/home_page.html
```diff
@@ -28,29 +27,6 @@ <h1 class="title">{{ page.page_title }}</h1>
...
-                </div>
-                {% endif %}
-
-                <div class="image__overlay ds-hero-image__overlay">
-                    <h1 class="title">{{ page.page_title }}</h1>
-                    <p>{{ page.page_subtitle }}</p>
-                </div>
-            </div>
-        </div>
-    </div>
```
philli-m, line 53: @hom3mad3 would like a second opinion on this, do you think it is better to have the above structure so display different HTML depending on desktop or mobile or to only implement the parlax only on large screens?

webpack.common.js
```diff
@@ -8,7 +8,6 @@ module.exports = {
   entry: {
     digitalstrategie_berlin: [
       './digitalstrategie/assets/berlin_css/berlin_marketing.css',
-      './digitalstrategie/assets/berlin_css/shariff.min.css',
       './digitalstrategie/assets/berlin_css/fontawesome.all.min.css'
```
philli-m, line 11: @goapunk I tried adding the problematic font files locally also however it still triggers the `downloadable font: rejected by sanitizer (font-family: "Font Awesome 5 Brands" style:normal weight:400 stretch:100 src index:1) source: http://localhost:8006/static/fonts/fa-brands-400.woff2 ` error and it's always the same extensions so I'm now thinking it is the CSP or corrupted files? chrome gives better errors, it also has 
`The source list for the Content Security Policy directive 'style-src-elem' contains an invalid source: 'https://www.berlin.dehttp://berlin.de'. It will be ignored.`
maybe we can check together?

philli-m: Added extra commit to just load it via tag as it's not needed locally really, right?

goapunk (approved):
LGTM

